PR #9, opened 2020-07-20
Add new failure scenarios

RenaudWasTaken:
Hello!

@adrianchiris you mentioned the following error:
> failure to apply any configuration specified in containerSpec by the runtime (devices, mounts, hooks)

Can you detail a bit more your thoughts?
Thanks!

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -193,3 +193,12 @@ The `containerSpec` field has the following definition:
 ## Error Handling
   * Two or more files with identical `kind` values and identical `containerRuntime` field.
     Container runtimes should surface this error when any device with that `kind` is requested.
+  * Kind requested is not present in any CDI file.
+    Container runtimes should surface an error when a non existent kind is requested.
+  * Device (not device node) Requested does not exist.
+    Container runtimes should surface this error when a non existent device is requested.
+  * "Resource" does not exist (e.g: Mount, Hook, ...).
+    Container runtimes should surface this error when a non existent "resource" is requested (e.g: at "run" time).
+    This is because a resource does not need to exist when the spec is written but it needs to exist when the container is created.
+  * Hook fails to execute.
+    Container runtimes should surface an error when hooks fails to execute.


SPEC.md
```diff
@@ -193,3 +193,7 @@ The `containerSpec` field has the following definition:
 ## Error Handling
   * Two or more files with identical `kind` values and identical `containerRuntime` field.
     Container runtimes should surface this error when any device with that `kind` is requested.
+  * Kind is not present in any CDI file.
+    Container runtimes should surface an error when a non existent kind is requested.
+  * Device Requested does not exist.
+    Container runtimes should surface this error when a non existend device is requested.
```
adrianchiris, line 199: existend -> existent

adrianchiris:
>Can you detail a bit more your thoughts?

So my thoughts are how would runtime behave if in a CDI specification `containerSpec`:
- a specified device file does not exists  or changing its mount namespace fails
- a specified mount does not exist or changing its mount namespace fails
- a specified hook does not exist or fails to execute  (can hooks fail ?)

RenaudWasTaken:
> a specified hook does not exist or fails to execute (can hooks fail ?)

Hooks can fail :D, they tend to give you an error such as this:
> Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:432: running prestart hook 1 caused \\\"error running hook: exit status 1, stdout: , stderr: ERROR MESSAGE \\\\n\\\"\"": unknown.


Adding the other failure mode!

RenaudWasTaken:
Addressed your comments @adrianchiris

SPEC.md
```diff
@@ -193,3 +193,12 @@ The `containerSpec` field has the following definition:
...
     Container runtimes should surface this error when any device with that `kind` is requested.
+  * Kind requested is not present in any CDI file.
+    Container runtimes should surface an error when a non existent kind is requested.
+  * Device (not device node) Requested does not exist.
+    Container runtimes should surface this error when a non existent device is requested.
+  * "Resource" does not exist (e.g: Mount, Hook, ...).
+    Container runtimes should surface this error when a non existent "resource" is requested (e.g: at "run" time).
+    This is because a resource does not need to exist when the spec is written but it needs to exist when the container is created.
+  * Hook fails to execute.
+    This is usually already handled by the runtime engine, it would return an error that the container runtime need to surface to the user.
```
adrianchiris, line 204: While its already handled today by (some) runtime engines, this bit reflects the expectations of this spec from container runtime in case of such error right?

IMO we should word this like the others e.g

> * Hook fails to execute.
   Container runtimes should surface an error when hooks fails to execute.

RenaudWasTaken: Fixed!

RenaudWasTaken:
@adrianchiris let me know if this is ok with you!

PR #12, opened 2020-07-23
Add Golang specs

RenaudWasTaken:
Add Golang specs to CDI

/cc @container-orchestrated-devices/cod 


RenaudWasTaken:
I'm not sure if the workflows can be viewed on my fork, I can however see a green CI here: https://github.com/RenaudWasTaken/container-device-interface/runs/908476002

SPEC.md
```diff
@@ -144,30 +144,30 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
...
 
-  * `cdiDevices` (array of objects, REQUIRED) list of devices provided by the vendor.
+  * `devices` (array of objects, REQUIRED) list of devices provided by the vendor.
     * `name` (string, REQUIRED), name of the device, can be used to refer to it when requesting a device.
       * Beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between.
       * e.g: `docker/podman run --device foo ...`
     * `nameShort` (array of strings, OPTIONAL), alternative names for the device. Can be used to reduce the CLI verbosity
       * Entries in the array MUST use the same schema as the entry for the `name` field
-    * `containerSpec` (object, OPTIONAL) this field is described in the next section.
+    * `ociEdits` (object, OPTIONAL) this field is described in the next section.
```
adrianchiris, line 156: I wonder if this spec should just specify what needs to happen to make a device available to a container.
OCI edits is the means the container runtime uses to satisfy this containerSpec.

i think the previous term `containerSpec` is more correct, as its notion is general and does not imply how a container runtime satisfies this spec.

we can say that when an OCI compliant runtime engine is used, this can be achieved through OCI edits.

SPEC.md
```diff
@@ -144,30 +144,30 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
...
       * This field should only be merged in the OCI spec if the device has been requested by the container runtime user.
 
 
-#### Container Spec
+#### OCI Edits
 
-The `containerSpec` field describes edits to be made to the OCI specification. Currently only three kinds of edits can be made to the OCI specification: `devices`, `mounts` and `hooks`.
+The `ociEdits` field describes edits to be made to the OCI specification. Currently only three kinds of edits can be made to the OCI specification: `devices`, `mounts` and `hooks`.
```
adrianchiris, line 162: Continuing from my comments above:
> The `ociEdits` field describes edits to be made to the OCI specification. Currently only three kinds of edits can be made to the OCI specification: `devices`, `mounts` and `hooks`.

would in turn be reworded to (or similar...):
The `containerSpec` field describes device nodes, mounts that need to be made available to containers as well as hooks that need to be executed at a particular time during the container creations stages in order to make a device available to the container.
when an OCI compliant container runtime engine is used this can be achieved by editing the OCI specification.

klueska:
It's hard to tell directly from the specs defined here, but two requirements I know we have for GPUs (especially ones with Mulit-Instance GPU support are):

1) An option for the runtime to call an executable that generates the CDI on the fly for a plugin, rather than simply reading a static file on disk that represents the CDI for a plugin. This will help in dynamic environments (such as MIG or any plug-and-play scenarios) where the specs might be changing frequently.

2) A spec for listing host files / directories “owned” by a plugin that should not be injected into a container by the runtime itself (e.g. /proc/drivers/nvidia).

RenaudWasTaken:
@klueska could you open an issue around these topics?

SPEC.md
RenaudWasTaken: I've been thinking about this too, for now we're looking at OCI edits, but it's definitely something we want to be different!
`ContainerSpec` is probably incorrect too, maybe `ContainerEdits` ?

adrianchiris: I see `containerSpec` as a generalization, where we specify what needs to be done in order to make a device available in a container. The container spec in this context would refer to CDI (limited to CDI) and not the "grand" container spec as represented by OCI.

If this sub-element (formally known as `containerSpec`) targets specifically OCI edits then its OK to keep it as is IMO.
We would probably be fine with it as long as OCI is the input a container runtime provides the container engine and edits to this spec is the only thing required.

bart0sh: +1 for keeping containerSpec name

specs-go/config.go
```diff
@@ -0,0 +1,49 @@
...
+	Hooks       []Hook       `json:"hooks,omitempty"`
+	Mounts      []Mount      `json:"mounts,omitempty"`
+}
+
+// DeviceNode represents a device node that needs to be added to the OCI spec.
+type DeviceNode struct {
+	HostPath      string `json:"hostPath"`
+	ContainerPath string `json:"containerPath"`
+	Permissions   string `json:"permissions,omitempty"`
+}
```
bart0sh, line 33: Would it make sense to use types from OCI runtime spec as much as possible?
For example we can use [LInuxDevice structure](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L377:L393) here.

The overall idea is to simplify updating OCI spec as far as I understood. Reusing spec structures would make it much easier I believe.

SPEC.md
RenaudWasTaken: Will update back to containerSpec

RenaudWasTaken: Edit: `ContainerEdits`

specs-go/config.go
RenaudWasTaken: Sorry for the very late reply, reusing the `LinuxDevice` structure has the unfortunate consequence that it makes it Linux specific. With `HostPath`/`ContainerPath` we keep this more or less OS agnostic.

SPEC.md
RenaudWasTaken: To me the spec really describe "what you should do" (you being the runtime) to make this device available to the container. Optionally, if you can't instruct the runtime with mounts, environment variables, ... you can specify hooks to do "what is needed" to make this device available to the container

SPEC.md
```diff
@@ -70,25 +70,25 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
     "cdiVersion": "0.1.0",
     "kind": "<name>",
     "kindShort": ["<short-name>", ...],
-    "container-runtime": ["<container-runtime-name>"], (optional)
+    "containerRuntime": ["<container-runtime-name>"], (optional)
```
mrunalp, line 73: why did we need this field for even though it is optional?

RenaudWasTaken: This field is here so that a vendor can have a CDI specification for specific runtimes. e.g: the nvidia-kata.json and the nvidia.json specification

specs-go/config.go
```diff
@@ -0,0 +1,49 @@
...
+	Name           string         `json:"name"`
+	NameShort      []string       `json:"nameShort"`
+	ContainerEdits ContainerEdits `json:"containerEdits"`
+}
+
+// ContainerEdits are edits a container runtime must make to the OCI spec to expose the device.
+type ContainerEdits struct {
+	DeviceNodes []DeviceNode `json:"deviceNodes,omitempty"`
+	Hooks       []Hook       `json:"hooks,omitempty"`
+	Mounts      []Mount      `json:"mounts,omitempty"`
```
mrunalp, line 25: Do we need an additional field to enforce ordering or we believe append would work?

RenaudWasTaken: For now we can probably stay on an append only structure and try to infer the ordering. If a field is required we can reconsider when the use case presents itself

Klaven (approved):
sorry late too the game, but this looks good to me.

PR #19, opened 2021-02-19
Add inital CDI code

RenaudWasTaken:
This Pull Request adds the initial code for CDI, we should then be able to import cdi/pkg from other projects with the following two functions:
- HasDevice -> Check if a Device given by a user is a CDI device
- UpdateOCISpecForDevices -> Update the OCI spec for a set of CDI device

This is missing tests for UpdateOCISpecForDevices but it's a first start.
This should also probably be optimized to cache the different files but these may come as a followup PR

### DIFF
--- a/pkg/devices.go
+++ b/pkg/devices.go
@@ -1,0 +1,180 @@
+package pkg
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	cdispec "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const (
+	root = "/etc/cdi"
+)
+
+func collectCDISpecs() (map[string]*cdispec.Spec, error) {
+	var files []string
+	vendor := make(map[string]*cdispec.Spec)
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if info == nil || info.IsDir() {
+			return nil
+		}
+
+		if filepath.Ext(path) != ".json" {
+			return nil
+		}
+
+		files = append(files, path)
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, path := range files {
+		spec, err := loadCDIFile(path)
+		if err != nil {
+			continue
+		}
+
+		if _, ok := vendor[spec.Kind]; ok {
+			continue
+		}
+
+		vendor[spec.Kind] = spec
+	}
+
+	return vendor, nil
+}
+
+// TODO: Validate (e.g: duplicate device names)
+func loadCDIFile(path string) (*cdispec.Spec, error) {
+	file, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	var spec *cdispec.Spec
+	err = json.Unmarshal([]byte(file), &spec)
+	if err != nil {
+		return nil, err
+	}
+
+	return spec, nil
+}
+
+/*
+* Pattern "vendor.com/device=myDevice" with the vendor being optional
+ */
+func extractVendor(dev string) (string, string) {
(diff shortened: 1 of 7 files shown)


RenaudWasTaken:
cc @zvonkok

pkg/devices.go
```diff
@@ -0,0 +1,210 @@
...
+	spec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	root = "/etc/cdi"
+)
+
+var (
+	defaultLogger = logrus.New()
```
mrunalp, line 21: We could avoid logging if we want to use this as a library. The callers could be returned errors and decide what to do.

pkg/devices.go
```diff
@@ -0,0 +1,210 @@
...
+	"path/filepath"
+	"strings"
+
+	cdispec "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	root = "/etc/cdi"
```
mrunalp, line 17: We can have a `New()`  function that takes this as an option while defaulting to this directory if not set.

pkg/devices.go
```diff
@@ -0,0 +1,210 @@
...
+
+		edits[spec.Kind] = spec
+		err = cdispec.ApplyOCIEditsForDevice(ociconfig, spec, d)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, spec := range edits {
+		cdispec.ApplyOCIEdits(ociconfig, spec)
```
kad, line 206: no error checking for those?

pkg/devices_test.go
```diff
@@ -0,0 +1,204 @@
+package pkg
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+
+	cdispec "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+	spec "github.com/opencontainers/runtime-spec/specs-go"
```
kad, line 8: unused import.

kad (changes requested):
at least error checking on `ApplyOCIEdits` needs to be fixed.

One other thing I've noticed, we somewhere in all discussions lost the ability for individual device instance to inject env variables into OCI spec. I need to recheck other places.

pkg/devices.go
kad, line 206:
```suggestion
        err := cdispec.ApplyOCIEdits(ociconfig, spec)
        if err != nil {
            return err
        }
```

pkg/devices.go
```diff
@@ -0,0 +1,210 @@
...
+	vendor := make(map[string]*cdispec.Spec)
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if info == nil || info.IsDir() {
+			log.Infof("CDI: Skipped directory %q", path)
+			return nil
+		}
+
+		if filepath.Ext(path) != ".json" {
+			log.Infof("CDI: Skipped non json terminated file %q", path)
```
uniemimu, line 39: Looks to me this is just printing, the append is still happening below

pkg/devices.go
RenaudWasTaken: Good catch!

pkg/devices_test.go
RenaudWasTaken: Good catch!

pkg/devices.go
RenaudWasTaken: Good catch!

RenaudWasTaken:
> One other thing I've noticed, we somewhere in all discussions lost the ability for individual device instance to inject env variables into OCI spec. I need to recheck other places.


Good comment, I'll make it part of the next PR

umohnani8:
Late to the party, but initial code LGTM :)

kad (approved):
LGTM.
Side note: we later probably want to switch to https://golangci-lint.run/ instead of classical golint, but that's not urgent.

Issue #20, opened 2021-03-13
Environment Variable setup


RenaudWasTaken:
What: The spec currently does not have the ability to edit the environment variables
Fix: Add Environment variables to the spec and the `ApplyEditsToOCISpec` function

bart0sh:
@RenaudWasTaken @kad Please review #21

kad:
@bart0sh as #19 and #21  are now merged, can you add one more PR to implement `env` inject into ApplyOCIEdits?

PR #21, opened 2021-03-18
add environment variables to the spec

bart0sh:


kad:
lgtm, with only one question to others: it looks like other fields in the spec are plurals. Should we also have `envs` instead of `env`? I'm fine with any of those, just curious on how others feel about that.

kad:
@bart0sh one more thing, that shouldn't be fixing #20 yet, as we need second part of the change to do in `ApplyOCIedits` in #19

bart0sh:
> it looks like other fields in the spec are plurals. Should we also have `envs` instead of `env`? I'm fine with any of those, just curious on how others feel about that.

It's called env in the OCI runtime spec, so I followed their naming convention.

bart0sh:
> @bart0sh one more thing, that shouldn't be fixing #20 yet, as we need second part of the change to do in `ApplyOCIedits` in #19

thanks for pointing out. removed.

SPEC.md
```diff
@@ -160,14 +163,15 @@ Note: For a CDI file to be valid, at least one entry must be specified in this a
...
-The `containerEdits` field describes edits to be made to the OCI specification. Currently only three kinds of edits can be made to the OCI specification: `devices`, `mounts` and `hooks`.
+The `containerEdits` field describes edits to be made to the OCI specification. Currently only four kinds of edits can be made to the OCI specification: `env`, `devices`, `mounts` and `hooks`.
 
 The `containerEdits` field is referenced in two places in the specification:
   * At the device level, where the edits MUST only be made if the matching device is requested by the container runtime user.
   * At the container level, where the edits MUST be made if any of the of the device defined in the `devices` field are requested.
 
 
 The `containerEdits` field has the following definition:
+  * `env` (array of strings in the format of "VARNAME=VARVALUE", OPTIONAL) describes the environment variables that should be set. These values act as defaults and are merged with any specified when creating a container.
```
RenaudWasTaken, line 174: Small nit we should specify that they appended (i.e: at the end) to the environment array

bart0sh: @RenaudWasTaken thank you for the review. I've updated the PR as you've suggested. Please, have a look.

RenaudWasTaken (approved):

Issue #17, opened 2020-09-22
uid and gid mappings missing


mythi:
CDI should specify how non-root uid/gid are set for the devices.

eero-t:
Otherwise Pods requesting devices need to run as root, which implies then having unnecessary access e.g. to device files in sysfs (which is mounted in its totality to containers).  I.e. this is a security issue.

RenaudWasTaken:
This is a request that makes sense, do you want to make a PR for it?

bart0sh:
I'd suggest to use the same approach as used in OCI runtime spec:
process/user contains the following fields:
- uid (int, REQUIRED) specifies the user ID in the container namespace.
- gid (int, REQUIRED) specifies the group ID in the container namespace.
- umask (int, OPTIONAL) specifies the [umask][umask_2] of the user. If unspecified, the umask should not be changed from the calling process' umask.
- additionalGids (array of ints, OPTIONAL) specifies additional group IDs in the container namespace to be added to the process.

if we want to cover Windows we can add username field:
- username (string, OPTIONAL) specifies the user name for the process

So, in CDI spec it would looks something like this:
```
"ociEdits": {
    "user": [
      {"uid": 1000, "gid": 1000}
    ],
```

@mythi @eero-t  would this cover your uid/gid requirements?

eero-t:
@bart0sh it's not clear to me from your comment context, is your suggestion for CDI internal API, or what user requesting a device will specify.  Latter is what this ticket is about, but "ociEdits" sounds more like former.  Could you clarify?

bart0sh:
It was a suggestion to use set of user related parameters from OCI runtime spec in the CDI spec.
I'm not sure I've answered your question though. Can you clarify?

RenaudWasTaken:
Cc @mrunalp @mikebrow want to have more input on this

RenaudWasTaken:
PR #27 just got merged - closing this

PR #24, opened 2021-03-25
add user to the spec

bart0sh:
Added 'User' object to the ociEdits section.
User fields and types were taken from the [OCI Runtime spec](https://github.com/opencontainers/runtime-spec/blob/master/config.md)

Fixes: #17

### DIFF
--- a/schema/testdata/good/spec-example.json
+++ b/schema/testdata/good/spec-example.json
@@ -17,6 +17,7 @@
       "FOO=CDI_SPEC",
       "BAR=BARVALUE1"
     ],
+    "user": {"uid": 1000, "gid": 1000, "umask": 7, "additionalGids": [2, 8]},
     "deviceNodes": [
       {"hostPath": "/dev/vendorctl", "containerPath": "/dev/vendorctl"}
     ],
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -21,6 +21,7 @@ type Devices struct {
 // ContainerEdits are edits a container runtime must make to the OCI spec to expose the device.
 type ContainerEdits struct {
 	Env         []string      `json:"env,omitempty"`
+	User        *User         `json:"user,omitempty"`
 	DeviceNodes []*DeviceNode `json:"deviceNodes,omitempty"`
 	Hooks       []*Hook       `json:"hooks,omitempty"`
 	Mounts      []*Mount      `json:"mounts,omitempty"`
@@ -31,6 +32,14 @@ type DeviceNode struct {
 	HostPath      string   `json:"hostPath"`
 	ContainerPath string   `json:"containerPath"`
 	Permissions   []string `json:"permissions,omitempty"`
+}
+
+// User represents a user the container process runs as
+type User struct {
+	UID            uint32   `json:"uid"`
+	GID            uint32   `json:"gid"`
+	Umask          uint32   `json:"umask,omitempty"`
+	AdditionalGids []uint32 `json:"additionalgids,omitempty"`
 }
 
 // Mount represents a mount that needs to be added to the OCI spec.
--- a/specs-go/oci.go
+++ b/specs-go/oci.go
@@ -2,6 +2,7 @@ package specs
 
 import (
 	"fmt"
+
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 )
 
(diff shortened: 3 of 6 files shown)


kad:
lgtm, however, I'm thinking if we really need to have uid/gid as mandatory fields. If those are not specified, we can always in theory copy from host.

RenaudWasTaken:
I'm not super sure about this one, it seems a bit odd that when I request a device I get a different uid/gid. I'm also not sure who would take precedence in the context where a user would specify a specific UID/GID.

What are the other options for this? I've lost context on why we arrived at this solution, sorry :(

kad:
@RenaudWasTaken the other context was on CRI->internal representation layer  where device ownership was originally copied from host, and nowadays we have patches that copies uid/gid from security context of Pod/Container from Kubernets.
For CLI usages, we don't have that... so we either need same algorithm of copying uid/gid from host (not very usable for non-root containers), or have that explicitly set in cdi.json (as @bart0sh  implemented) and/or do some smart uid/gid mapping from OCI spec if not explicitly stated in cdi.json. 

I'm more inclined to this 3rd method:
- if uid or gid fields are not explicitly specified in cdi.json - try to get those from uid/gid of OCI spec of container.
- if for some reason we can't do that, copy uid/gid from host os.

bart0sh:
@kad >  I'm thinking if we really need to have uid/gid as mandatory fields

would it make sense to only set uid/gid if they're not set in OCI config?

kad:
> @kad > I'm thinking if we really need to have uid/gid as mandatory fields
> 
> would it make sense to only set uid/gid if they're not set in OCI config?

I think we should more mimicking the behaviour of CRI-O/containerd -> when runtimes got information about device from CRI, they don't have enough information to construct proper OCI part of the spec for it. Thus they do some "copy from host" things or doing some defaulting for the fields. IMHO, we should do same here: if we got from cdi.json enough information -> add to OCI spec. If we end-up that some mandatory by OCI fields are missing - we should be able to fill them with some reasonable defaults.

bart0sh:
> If we end-up that some mandatory by OCI fields are missing - we should be able to fill them with some reasonable defaults.

We should probably define when cdi.json should be processed by CRI runtimes. If it's processed after runtimes set user properties then there is no need to duplicate filling user properties with defaults by CDI. In this case we would only need to set them if cdi.json has them but the OCI config doesn't. This way we ensure we don't rewrite them.

specs-go/config.go
```diff
@@ -33,6 +34,14 @@ type DeviceNode struct {
 	Permissions   []string `json:"permissions,omitempty"`
 }
 
+// User represents a user the container process runs as
+type User struct {
+	UID            uint32   `json:"uid"`
+	GID            uint32   `json:"gid"`
+	Umask          uint32   `json:"umask,omitempty"`
+	AdditionalGids []uint32 `json:"additionalgids,omitempty"`
+}
```
mikebrow, line 43: What do you think about using pointers for unset on omit empty...

specs-go/oci.go
```diff
@@ -31,13 +32,15 @@ func ApplyEditsToOCISpec(config *spec.Spec, edits *ContainerEdits) error {
...
-		if config.Process == nil {
-			config.Process = &spec.Process{}
+	if config.Process == nil {
+		fmt.Println("CDI: config: process is not defined")
+	} else {
+		if len(edits.Env) > 0 {
+			config.Process.Env = append(config.Process.Env, edits.Env...)
+		}
+		if edits.User != nil {
+			config.Process.User = spec.User{UID: edits.User.UID, GID: edits.User.GID, Umask: edits.User.Umask, AdditionalGids: edits.User.AdditionalGids}
```
mikebrow, line 42: see above

schema/testdata/good/spec-example.json
```diff
@@ -17,6 +17,7 @@
       "FOO=CDI_SPEC",
       "BAR=BARVALUE1"
     ],
+    "user": {"uid": 1000, "gid": 1000, "umask": 7, "additionalGids": [2, 8]},
```
mikebrow, line 20: kk and I think we need some additional spec-examples.. or modify the basic test case for handling omitempty variants in test..

mikebrow (commented):
see comments

RenaudWasTaken:
I'm also not super clear on how we handle the fact that a user may have already specified a UID/GID.
Whether through the CLI or through the kubernetes securityContext fields

specs-go/config.go
```diff
@@ -33,6 +34,14 @@ type DeviceNode struct {
 	Permissions   []string `json:"permissions,omitempty"`
 }
 
+// User represents a user the container process runs as
+type User struct {
+	UID            uint32   `json:"uid"`
+	GID            uint32   `json:"gid"`
```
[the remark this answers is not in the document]
bart0sh: I used types from the [OCI spec](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#Devices)

specs-go/config.go
bart0sh: that makes sense. Will do.

bart0sh:
I'm going to close this PR and come up with a new one. This one targets OCI spec 'Process', which is incorrect. UID, GID and other properties(type, major, minor, etc) should be set for the [Linux.Devices](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#Devices)

Issue #26, opened 2021-03-30
CDI spec doesn't provide enough info about devices


bart0sh:
CDI spec doesn't have the following device properties that present in the [OCI Linux.Devices spec](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#Devices):
- device type
- major
- minor
- fileMode
- uid
- gid

And doesn't target Linux.Devices at all in its current implementation.

bart0sh:
@RenaudWasTaken @mrunalp To address this issue we should add missing properties to the DeviceNode and use them to populate Linux.Devices in OCI config. Does this make sense to you?

RenaudWasTaken:
> @RenaudWasTaken @mrunalp To address this issue we should add missing properties to the DeviceNode and use them to populate Linux.Devices in OCI config. Does this make sense to you?

Completely, I was looking through some of our discussions to understand why we weren't going down that route  :D !

RenaudWasTaken:
We can do that :) !

elezar:
@bart0sh as mentioned in the meeting today, when I was experimenting with direct OCI modifications to get Devices working, I needed access to the device major / minor numbers to update the cgroups in the `LinuxResources`.

There are also devices (e.g. MIG devices) for which the device nodes may not exist on the host.

PR #27, opened 2021-04-28
extend DeviceNode definition

bart0sh:
Added type, major, minor, uid and gid properties to the DeviceNode definition.
Property types were taken from the [OCI LinuxDevice](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/specs-go/config.go#L377:L393).

Fixes: #26, #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ $ cat > /etc/cdi/vendor.json <<EOF
       "name": "myDevice",
       "ociEdits": {
         "deviceNodes": [
-          {"hostPath": "/dev/card1", "containerPath": "/dev/card1"}
-          {"hostPath": "/dev/card-render1", "containerPath": "/dev/card-render1"}
+          {"path": "/dev/card1", type": "c", "major": 25, "minor": 25, "permissions": "rw", "uid": 1000, "gid": 1000}
+          {"path": "/dev/card-render1", "type": "c", "major": 25, "minor": 25, "permissions": "rwm", "uid": 1000, "gid": 1000}
         ]
       }
     }
@@ -44,7 +44,7 @@ $ cat > /etc/cdi/vendor.json <<EOF
       "BAR=BARVALUE1"
     ],
     "deviceNodes": [
-      {"hostPath": "/dev/vendorctl", "containerPath": "/dev/vendorctl"}
+      {"path": "/dev/vendorctl", "type": "b", "major": 25, "minor": 25, "permissions": "rw", "uid": 1000, "gid": 1000}
     ],
     "mounts": [
       {"hostPath": "/bin/vendorBin", "containerPath": "/bin/vendorBin"},
--- a/SPEC.md
+++ b/SPEC.md
@@ -94,14 +94,17 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
             ]
             "deviceNodes": [ (optional)
                 {
-                    "hostPath": "<path>",
-                    "containerPath": "<path>",
-
+                    "path": "<path>",
+                    "type": "<type>" (optional),
+                    "major": <int32> (optional),
+                    "minor": <int32> (optional),
                     // Cgroups permissions of the device, candidates are one or more of
                     // * r - allows container to read from the specified device.
                     // * w - allows container to write to the specified device.
                     // * m - allows container to create device files that do not yet exist.
-                    "permissions": "<permissions>" (optional)
+                    "permissions": "<permissions>" (optional),
+                    "uid": <int> (optional),
+                    "gid": <int> (optional)
                 }
             ]
             "mounts": [ (optional)
@@ -163,7 +166,7 @@ Note: For a CDI file to be valid, at least one entry must be specified in this a
 
 #### OCI Edits
 
-The `containerEdits` field describes edits to be made to the OCI specification. Currently only four kinds of edits can be made to the OCI specification: `env`, `devices`, `mounts` and `hooks`.
+The `containerEdits` field describes edits to be made to the OCI specification. Currently the following kinds of edits can be made to the OCI specification: `env`, `devices`, `mounts` and `hooks`.
 
 The `containerEdits` field is referenced in two places in the specification:
   * At the device level, where the edits MUST only be made if the matching device is requested by the container runtime user.
@@ -173,12 +176,16 @@ The `containerEdits` field is referenced in two places in the specification:
 The `containerEdits` field has the following definition:
   * `env` (array of strings in the format of "VARNAME=VARVALUE", OPTIONAL) describes the environment variables that should be set. These values are appended to the container environment array.
   * `deviceNodes` (array of objects, OPTIONAL) describes the device nodes that should be mounted:
-    * `hostPath` (string, REQUIRED) path of the device on the host.
-    * `containerPath` (string, REQUIRED) path of the device within the container.
+    * `path` (string, REQUIRED) path of the device within the container.
+    * `type` (string, OPTIONAL) Device type: block, char, etc.
+    * `major` (int64, OPTIONAL) Device major number.
+    * `minor` (int64, OPTIONAL) Device minor number.
     * `permissions` (string, OPTIONAL) Cgroups permissions of the device, candidates are one or more of:
       * r - allows container to read from the specified device.
       * w - allows container to write to the specified device.
       * m - allows container to create device files that do not yet exist.
+    * `uid` (uint32, OPTIONAL) id of device owner in the container namespace.
+    * `gid` (uint32, OPTIONAL) id of device group in the container namespace.
   * `mounts` (array of objects, OPTIONAL) describes the mounts that should be mounted:
     * `hostPath` (string, REQUIRED) path of the device on the host.
     * `containerPath` (string, REQUIRED) path of the device within the container.
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -6,6 +6,11 @@
             "minimum": 0,
             "maximum": 4294967295
         },
+        "int64": {
+            "type": "integer",
+            "minimum": -9223372036854775808,
+            "maximum": 9223372036854775807
+        },
         "ArrayOfStrings": {
             "type": "array",
             "items": {
@@ -21,19 +26,30 @@
         "DeviceNode": {
             "type": "object",
             "properties": {
-                "hostPath": {
-                    "$ref": "#/definitions/FilePath"
-                },
-                "containerPath": {
+                "path": {
                     "$ref": "#/definitions/FilePath"
                 },
                 "permissions": {
                     "type": "string"
+                },
+                "type": {
+                    "type": "string"
+                },
+                "major": {
+                    "$ref": "#/definitions/int64"
+                },
+                "minor": {
+                    "$ref": "#/definitions/int64"
+                },
+                "uid": {
+                    "$ref": "#/definitions/uint32"
+                },
+                "gid": {
+                    "$ref": "#/definitions/uint32"
                 }
             },
             "required": [
-                "hostPath",
-                "containerPath"
+                "path"
             ]
         },
         "Mount": {
--- a/schema/testdata/good/spec-example.json
+++ b/schema/testdata/good/spec-example.json
@@ -6,8 +6,8 @@
       "name": "myDevice",
       "ociEdits": {
         "deviceNodes": [
-          {"hostPath": "/dev/card1", "containerPath": "/dev/card1"},
-          {"hostPath": "/dev/card2", "containerPath": "/dev/card2"}
+          {"path": "/dev/card1"},
+          {"path": "/dev/card2"}
         ]
       }
     }
@@ -18,7 +18,7 @@
       "BAR=BARVALUE1"
     ],
     "deviceNodes": [
-      {"hostPath": "/dev/vendorctl", "containerPath": "/dev/vendorctl"}
+      {"path": "/dev/vendorctl", "uid": 1000, "gid": 1000}
     ],
     "mounts": [
       {"hostPath": "/bin/vendorBin", "containerPath": "/bin/vendorBin"},
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -28,9 +28,13 @@ type ContainerEdits struct {
 
 // DeviceNode represents a device node that needs to be added to the OCI spec.
 type DeviceNode struct {
-	HostPath      string   `json:"hostPath"`
-	ContainerPath string   `json:"containerPath"`
-	Permissions   []string `json:"permissions,omitempty"`
+	Path        string   `json:"path"`
+	Type        string   `json:"type,omitempty"`
+	Major       int64    `json:"major,omitempty"`
+	Minor       int64    `json:"minor,omitempty"`
+	Permissions []string `json:"permissions,omitempty"`
+	UID         *uint32  `json:"uid,omitempty"`
+	GID         *uint32  `json:"gid,omitempty"`
 }
 
 // Mount represents a mount that needs to be added to the OCI spec.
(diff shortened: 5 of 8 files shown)


bart0sh:
@RenaudWasTaken @mrunalp  @kad does this look reasonable? Any concerns or suggestions? If it looks OK I'll update the tests and submit it for the final review.

SPEC.md
```diff
@@ -101,7 +101,9 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
                     // * r - allows container to read from the specified device.
                     // * w - allows container to write to the specified device.
                     // * m - allows container to create device files that do not yet exist.
-                    "permissions": "<permissions>" (optional)
+                    "permissions": "<permissions>" (optional),
+                    "uid": <int> (optional),
+                    "gid": <int> (optional)
```
elezar, line 106: Do we expect this to also contain the major, minor, and type?

README.md
```diff
@@ -32,8 +32,8 @@ $ cat > /etc/cdi/vendor.json <<EOF
       "name": "myDevice",
       "ociEdits": {
         "deviceNodes": [
-          {"hostPath": "/dev/card1", "containerPath": "/dev/card1"}
-          {"hostPath": "/dev/card-render1", "containerPath": "/dev/card-render1"}
+          {"path": "/dev/card1", "type": "c", "major": 25, "minor": 25, "permissions": "rw", "uid": 1000, "gid": 1000}
```
elezar, line 35: The code seems to indicate that this is still `hostPath`. Should that be updated there too? Do we have to support both?

specs-go/config.go
```diff
@@ -30,7 +30,12 @@ type ContainerEdits struct {
 type DeviceNode struct {
 	HostPath      string   `json:"hostPath"`
 	ContainerPath string   `json:"containerPath"`
+	Type          string   `json:"type"`
+	Major         int64    `json:"major"`
+	Minor         int64    `json:"minor"`
 	Permissions   []string `json:"permissions,omitempty"`
+	UID           *uint32  `json:"uid,omitempty"`
+	GID           *uint32  `json:"umask,omitempty"`
```
elezar, line 38: The json field name is incorrect. Should be `gid`. Also, should we drop the pointer specifier here? (not sure how json decoding works in this case).

specs-go/config.go
```diff
@@ -30,7 +30,12 @@ type ContainerEdits struct {
 type DeviceNode struct {
 	HostPath      string   `json:"hostPath"`
 	ContainerPath string   `json:"containerPath"`
+	Type          string   `json:"type"`
```
elezar, line 33: Should we add `omitempty` to the optional fields?

specs-go/oci.go
```diff
@@ -102,3 +101,14 @@ func toOCIDevice(d *DeviceNode) spec.Mount {
 		Options:     d.Permissions,
 	}
 }
+
+func toOCILinuxDevice(d *DeviceNode) spec.LinuxDevice {
```
elezar, line 105: Possibly out of scope: Do we also want to modify the cgroups here?

specs-go/oci.go
```diff
@@ -37,15 +37,14 @@ func ApplyEditsToOCISpec(config *spec.Spec, edits *ContainerEdits) error {
...
 		if config.Process == nil {
 			config.Process = &spec.Process{}
 		}
-
 		config.Process.Env = append(config.Process.Env, edits.Env...)
 	}
 
 	for _, d := range edits.DeviceNodes {
+		config.Linux.Devices = append(config.Linux.Devices, toOCILinuxDevice(d))
 		config.Mounts = append(config.Mounts, toOCIDevice(d))
```
elezar, line 48: Should we remove the mounts? In my testing, it was not needed to mount the devices if they were specified as `LinuxDevices` along with their cgroups.

elezar (commented):
Thanks @bart0sh. I have some minor comments / questions.

One additional question: Does using `LinuxDevice` as the base for how we define our devices limit our implementation to linux environments? Is there a desire to have this work on Windows at some point, for example?

SPEC.md
```diff
@@ -101,7 +101,9 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
                     // * r - allows container to read from the specified device.
                     // * w - allows container to write to the specified device.
                     // * m - allows container to create device files that do not yet exist.
-                    "permissions": "<permissions>" (optional)
+                    "permissions": "<permissions>" (optional),
```
RenaudWasTaken, line 104: Line 97 and 98 (`hostPath": "<path>",  "containerPath": "<path>",`) needs updating

SPEC.md
```diff
@@ -101,7 +101,9 @@ The key words "must", "must not", "required", "shall", "shall not", "should", "s
...
+                    "uid": <int> (optional),
```
RenaudWasTaken, line 105: need to add major and minor

SPEC.md
RenaudWasTaken: +1

SPEC.md
```diff
@@ -175,10 +177,15 @@ The `containerEdits` field has the following definition:
   * `deviceNodes` (array of objects, OPTIONAL) describes the device nodes that should be mounted:
     * `hostPath` (string, REQUIRED) path of the device on the host.
```
RenaudWasTaken, line 178: Needs updating (hostPath + containerPath)

schema/defs.json
```diff
@@ -29,6 +34,21 @@
                 },
                 "permissions": {
                     "type": "string"
+                },
```
RenaudWasTaken, line 37: hostPath + containerPath needs updating

schema/testdata/good/spec-example.json
```diff
@@ -18,7 +18,7 @@
       "BAR=BARVALUE1"
     ],
     "deviceNodes": [
-      {"hostPath": "/dev/vendorctl", "containerPath": "/dev/vendorctl"}
+      {"hostPath": "/dev/vendorctl", "containerPath": "/dev/vendorctl", "uid": 1000, "gid": 1000}
```
RenaudWasTaken, line 21: Needs to be updated

specs-go/oci.go
RenaudWasTaken: +1

specs-go/oci.go
RenaudWasTaken: @elezar I think cgroups are up to the runtime to setup

specs-go/config.go
naterini: Are we sure we want to use the numeric uid/gid here and not the alphanumeric instead? There should be no problem to resolve such username/groupname is given otherwise I would consider that an error at startup.

RenaudWasTaken (commented):
Generally looks good - don't need to respect backwards compat we haven't published G.A yet

specs-go/config.go
bart0sh: I was trying to be as close as possible to the [target structure](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/specs-go/config.go#L377:L393).
Alphabetic uids can be distro-specific. Not sure we want that.

SPEC.md
bart0sh: I haven't changed hostPath and containerPath. I've only added new fields.

SPEC.md
bart0sh: same thing here.
BTW, do you think it makes sense to get rid of hostPath and containerPath in favour of path?

bart0sh:
@elezar @RenaudWasTaken thank you for the review. I've updated the PR according to your comments.

SPEC.md
bart0sh: done

README.md
bart0sh: It was a mistake. I had no intantion to change host* attributes. Should be fixed now.

specs-go/config.go
bart0sh: done

specs-go/oci.go
bart0sh: done

SPEC.md
bart0sh: done

SPEC.md
elezar: Is there any use case for mounting a device at a different path in the container?

specs-go/oci.go
elezar: In my testing, I had to explicitly modify the `spec.Linux.Resources` appending `LinuxDeviceCgroup` entries as in the following example (where the `Inject` method of the `deviceNode` is responsible for modifying the spec):

```golang
func (d deviceNode) Inject(spec *oci.Spec) error {
	if spec.Linux == nil {
		spec.Linux = &oci.Linux{}
	}
	if spec.Linux.Resources == nil {
		spec.Linux.Resources = &oci.LinuxResources{}
	}

	// TODO: These need to be configurable
	deviceFileMode := os.FileMode(8630)
	deviceUID := uint32(0)
	deviceGID := uint32(0)

	deviceMajor := int64(d.Major)
	deviceMinor := int64(d.Minor)

	ociDevice := oci.LinuxDevice{
		Path:     string(d.Path),
		Type:     "c",
		Major:    deviceMajor,
		Minor:    deviceMinor,
		FileMode: &deviceFileMode,
		UID:      &deviceUID,
		GID:      &deviceGID,
	}
	spec.Linux.Devices = append(spec.Linux.Devices, ociDevice)

	ociDeviceCgroup := oci.LinuxDeviceCgroup{
		Allow:  true,
		Type:   "c",
		Major:  &deviceMajor,
		Minor:  &deviceMinor,
		Access: "rwm",
	}

	spec.Linux.Resources.Devices = append(spec.Linux.Resources.Devices, ociDeviceCgroup)

	return nil
}
```
(I know that there are some pieces that are not quite plumbed through there).

The runtime still sets up the cgroups using these resource specifications.

SPEC.md
elezar: I think @RenaudWasTaken was suggesting that we drop these in favour of a single `path` while we are in a position to make breaking changes.

bart0sh: This should be part of another PR I think.

SPEC.md
bart0sh: I think there was a reason. Otherwise hostPath and containerPath would not be used in the current spec. I don't know what is it, however. @RenaudWasTaken should know for sure.

SPEC.md
bart0sh: We can also agree to drop Permissions if favor of FileMode. This way our structure would be identical to the OCI LinuxDevice.

README.md
```diff
@@ -32,8 +32,8 @@ $ cat > /etc/cdi/vendor.json <<EOF
       "name": "myDevice",
       "ociEdits": {
         "deviceNodes": [
-          {"hostPath": "/dev/card1", "containerPath": "/dev/card1"}
-          {"hostPath": "/dev/card-render1", "containerPath": "/dev/card-render1"}
+          {"hostPath": "/dev/card1", "containerPath": "/dev/card1", type": "c", "major": 25, "minor": 25, "permissions": "rw", "uid": 1000, "gid": 1000}
+          {"hostPath": "/dev/card-render1", "contaierPath": "/dev/card-render1", "type": "c", "major": 25, "minor": 25, "permissions": "rwm", "uid": 1000, "gid": 1000}
```
mythi, line 36: small typo here

schema/defs.json
```diff
@@ -29,6 +34,21 @@
...
+                "type": {
+                    "type": "string"
+                },
+                "major": {
+                    "$ref": "#/devinitions/int64"
```
mythi, line 42: definitions (for the next 3 items below too)

SPEC.md
RenaudWasTaken: +1 to that

SPEC.md
RenaudWasTaken: With the edits that you are making - "hostPath" and "containerPath" no longer make any sense.

The structure basically describes a deviceNode that must be created within the container. In that definition of the structure the name "path" makes a lot of sense because it will never refer to a path on the host.

specs-go/oci.go
RenaudWasTaken: Oh - this seems so obvious now that you bring it up :D !
I think it makes sense to have it here - I agree that this should be up to the CDI code (and not spec) to take care.

I'll leave it up to @bart0sh on whether he wants to make it a followup change or part of this PR though.

SPEC.md
RenaudWasTaken: To clarify I'm in favor of removing `hostPath` and `containerPath` in favor of `path`, in the `deviceNodes` structure, in this PR because this PR changes the definition of that structure, from a simple "mount" instruction to a "create Device" instruction.

It seems reasonable at this point to rename the fields that are going to be changed because of that redifinition.

specs-go/config.go
```diff
@@ -30,7 +30,12 @@ type ContainerEdits struct {
 type DeviceNode struct {
 	HostPath      string   `json:"hostPath"`
 	ContainerPath string   `json:"containerPath"`
+	Type          string   `json:"type,omitempty"`
+	Major         int64    `json:"major,omitempty"`
+	Minor         int64    `json:"minor,omitempty"`
 	Permissions   []string `json:"permissions,omitempty"`
+	UID           *uint32  `json:"uid,omitempty"`
+	GID           *uint32  `json:"git,omitempty"`
```
elezar, line 38: typo in label: `git` -> `gid`

specs-go/oci.go
```diff
@@ -102,3 +100,14 @@ func toOCIDevice(d *DeviceNode) spec.Mount {
...
 }
+
+func toOCILinuxDevice(d *DeviceNode) spec.LinuxDevice {
+	return spec.LinuxDevice{
+		Path:  d.ContainerPath,
+		Type:  d.Type,
+		Major: d.Major,
+		Minor: d.Minor,
+		UID:   d.UID,
+		GID:   d.GID,
```
elezar, line 111: We don't set `FileMode` based on the contents of permission here. I assume that this will be addressed in a follow-up PR.

specs-go/oci.go
bart0sh: I'd prefer to make it a follow up change.

SPEC.md
bart0sh: done

README.md
bart0sh: done

schema/defs.json
bart0sh: done

specs-go/oci.go
bart0sh: correct

specs-go/config.go
bart0sh: fixed

bart0sh:
> Thanks @bart0sh. I have some minor comments / questions.
> 
> One additional question: Does using `LinuxDevice` as the base for how we define our devices limit our implementation to linux environments? Is there a desire to have this work on Windows at some point, for example?

To support Windows we'd need to add [Windows Device properties](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/specs-go/config.go#L496:L502): ID and IDType, make Path optional and change oci.go code to detect the OS type and use either LinuxDevice or WindowsDevice as a target. However, I'm not sure we need to do that in this PR. @RenaudWasTaken @kad WDYT?

schema/defs.json
bart0sh: done

specs-go/oci.go
elezar: A follow-up is fine.

RenaudWasTaken:
> To support Windows we'd need to add Windows Device properties: ID and IDType, make Path optional and change oci.go code to detect OS type and use either LinuxDevice or WindowsDevice as a target. However, I'm not sure we need to do that in this PR. @RenaudWasTaken @kad WDYT?


Windows support can be added as a followup :) !

RenaudWasTaken (approved):

PR #31, opened 2021-06-30
Remove unused toOCIDevice function

elezar:
This change removes an unused toOCIDevice function that was converting a
CDI Device node to an OCI Mount. This is no longer required as the CDI devices
are now represented as OCI Linux Devices.


elezar:
@RenaudWasTaken I think this was the one function that we mentioned in our discussion yesterday.

RenaudWasTaken (approved):

PR #32, opened 2021-06-30
Rename specs.Devices to specs.Device

elezar:
This change renames the type specs.Devices to specs.Device as the intent
is to represent a single device. This improves the redability of the
Devices slice which how has the type []Device and not []Devices.

### DIFF
--- a/pkg/devices_test.go
+++ b/pkg/devices_test.go
@@ -1,12 +1,12 @@
 package pkg
 
 import (
-	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


elezar:
cc @RenaudWasTaken 

Note, this is probably a breaking change for integration with Podman. I can create a follow-up PR to do the rename there too (possibly waiting for the permission corrections from @bart0sh).

RenaudWasTaken (approved):

PR #40, opened 2021-12-03
specs-go, schema: remove KindShort and NameShort.

klihub:
Remove unused Spec.KindShort and Device.NameShort, as agreed in the last COD gathering.


elezar (changes requested):
These are also referenced in `SPEC.md`. We should remove them from there too for consistency.

elezar (approved):
Thanks. Looks good now.

PR #42, opened 2022-01-17
Makefile: simplify fmt, lint, and vet targets.

klihub:


Makefile
```diff
@@ -37,7 +37,7 @@ pre-pr-checks pr-checks: test fmt lint vet
 
 
 fmt format:
-	$(Q)report=$$($(GO_FMT) -s -d -w $$(find . -name *.go)); \
+	$(Q)report=$$($(GO_FMT) -s -d -w $$(find . -name \*.go)); \
```
elezar, line 40: Is there a reason that one doesn't use `go list` instead of `find`? It is my understanding that this would exclude vendored dependencies, for example.

klihub: Probably no particular reason, just happened to be copy-pasted in its current form. I check how to best update each of these rules and update this PR.

elezar: For the record, it's not a blocker for this PR. If you think it's too much effort, we can keep it as is since it's at least locally consistent. We can then reevaluate at a later stage.

elezar (approved):

PR #43, opened 2022-01-24
devel: better support for runtime integration

klihub:
This PR improves support for runtime CDI integration. See this accompanying branch for [cri-o CDI integration](https://github.com/klihub/cri-o/tree/devel/cdi-integration) as an example of using it.

In particular this PR
  - improves CDI device validation
  - improves CDI device injection to be more in line with runtime expectations
  - adds support for annotating and extracting CDI devices from annotations

Requesting injection via annotation is a stop-gap solution until the CRI protocol gains dedicated CDI support.


kad:
from my perspective, looks good.

bart0sh:
I've tested this on both sides - in the device plugin and in containerd. Worked just fine.
@elezar please review & merge

elezar:
Hi sorry. Been off sick. Will have a look once I'm back.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+			default:
+				return "", errors.Errorf("invalid name %q, invalid charcter '%c'",
+					name, c)
+			}
+		}
+	}
+	if c := rune(name[len(name)-1]); !isAlphaNumeric(c) {
+		return "", errors.Errorf("invalid name %q, last '%c' should be alphanumeric",
+			name, c)
+	}
```
elezar, line 119: Is this not just the same as `ValidateVendorName` (minus the check for an empty string)? Does it make sense to factor this out into `validateAnnotationKeySuffix` or something (note that it's private)?

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+	value, sep := "", ""
+	for _, d := range devices {
+		if _, _, _, err := ParseQualifiedName(d); err != nil {
+			return "", err
+		}
+		value += sep + d
+		sep = ","
+	}
+
+	return value, nil
```
elezar, line 135: What about:
```
        var valid []string
	for _, d := range devices {
		if _, _, _, err := ParseQualifiedName(d); err != nil {
			return "", err
		}
		valid = append(valid, d)
	}

	return strings.Join(valid, ","), nil
```

Note, we could also just call `IsQualifiedName` if we don't think that the additional information in the error is useful.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+	const maxNameLen = 63
+
+	if plugin == "" {
+		return "", errors.New("invalid plugin name, empty")
+	}
+	if deviceID == "" {
+		return "", errors.New("invalid deviceID, empty")
+	}
+
+	name := plugin + "_" + strings.ReplaceAll(deviceID, "/", "_")
```
elezar, line 95: Question: Is there a specific reason that we don't allow `/` in the key? Does replacing them here not make debugging failures more difficult as a user would have to mentally perform this replacement?

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+			case isAlphaNumeric(c):
+			case c == '_' || c == '-' || c == '.':
+			default:
+				return "", errors.Errorf("invalid name %q, invalid charcter '%c'",
+					name, c)
+			}
+		}
+	}
+	if c := rune(name[len(name)-1]); !isAlphaNumeric(c) {
+		return "", errors.Errorf("invalid name %q, last '%c' should be alphanumeric",
```
elezar, line 117: Why is this a requirement? Is the only hard requirement not that the string cannot end on (or contain) `=` since this is used for the key-value representation of the annotations?

elezar: I would argue that we need to be as unrestrictive as possible here. The motivation being that we are looking to support multiple plugin vendors / implementations which may have vastly different ways of representing device IDs and / or plugin names. Requiring that these change or that different "variants" be tracked seems like unnecessary additional overhead.

pkg/cdi/annotations_test.go
```diff
@@ -0,0 +1,458 @@
...
+					devID:  "device",
+					devices: []string{
+						"vendor.com/class=device1",
+						"vendor.com/class=device2",
+						"vendor.com/class=device3",
+					},
+				},
+			},
+			annotations: map[string]string{
+				AnnotationPrefix + "vendor.class_device": "vendor.com/class=device1,vendor.com/class=device2,vendor.com/class=device3",
```
elezar, line 156: Nit: I know it's useful to add `AnnotationPrefix` here, but it would also be nice to see in the tests how the full annotation key will look and as such it may be useful to explicitly add the `cdi.k8s.io/` for the tests.

pkg/cdi/container-edits.go
```diff
@@ -138,6 +175,27 @@ func (e *ContainerEdits) Validate() error {
 	return nil
 }
 
+// Append other edits into this one. If called with a nil receiver,
+// allocates and returns newly allocated edits.
+func (e *ContainerEdits) Append(o *ContainerEdits) *ContainerEdits {
+	if o == nil || o.ContainerEdits == nil {
+		return e
```
elezar, line 182: This return does not allocate a new edit if the receiver is `nil`. Is that a concern?

pkg/cdi/container-edits.go
```diff
@@ -138,6 +175,27 @@ func (e *ContainerEdits) Validate() error {
...
+		e = &ContainerEdits{}
+	}
+	if e.ContainerEdits == nil {
+		e.ContainerEdits = &specs.ContainerEdits{}
+	}
+
+	e.Env = append(e.Env, o.Env...)
+	e.DeviceNodes = append(e.DeviceNodes, o.DeviceNodes...)
+	e.Hooks = append(e.Hooks, o.Hooks...)
+	e.Mounts = append(e.Mounts, o.Mounts...)
```
elezar, line 194: Question: Should this care about duplication?

pkg/cdi/container-edits.go
```diff
@@ -78,12 +83,44 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 	if len(e.Env) > 0 {
 		specgen.AddMultipleProcessEnv(e.Env)
 	}
+
 	for _, d := range e.DeviceNodes {
-		specgen.AddDevice(d.ToOCI())
+		dev := d.ToOCI()
+		if err := fillMissingInfo(&dev); err != nil {
```
elezar, line 89: Should we always fill the missing information? Are there not use cases where we do not want to query the devices -- for example if the spec is already complete?

pkg/cdi/container-edits.go
```diff
@@ -220,3 +286,72 @@ func ensureOCIHooks(spec *oci.Spec) {
 		spec.Hooks = &oci.Hooks{}
 	}
 }
+
+// fillMissingInfo fills in missing mandatory attributes from the host device.
+func fillMissingInfo(dev *oci.LinuxDevice) error {
+	if dev.Type != "" && (dev.Major != 0 || dev.Type == "p") {
+		return nil
+	}
+	hostDev, err := runc.DeviceFromPath(dev.Path, "rwm")
```
elezar, line 295: Does this mean that we require `rwm` permissions on the host device?

pkg/cdi/container-edits.go
```diff
@@ -220,3 +286,72 @@ func ensureOCIHooks(spec *oci.Spec) {
 		spec.Hooks = &oci.Hooks{}
 	}
 }
+
+// fillMissingInfo fills in missing mandatory attributes from the host device.
+func fillMissingInfo(dev *oci.LinuxDevice) error {
+	if dev.Type != "" && (dev.Major != 0 || dev.Type == "p") {
```
elezar, line 292: So if the device type and major are set we return. Should we also check the minor number?

elezar (commented):
Thanks @klihub. This looks good.

I have some minor comments regarding naming and validation. In general this looks good though.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+// are returned along with a non-nil error. The annotations are expected
+// to be formatted by, or in a compatible fashion to AnnotateInjection().
+func ParseAnnotations(annotations map[string]string) ([]string, []string, error) {
+	var (
+		keys    []string
+		devices []string
+	)
+
+	for key, value := range annotations {
+		if strings.HasPrefix(key, AnnotationPrefix) {
```
elezar, line 68: Maybe a nit: checking the logical no of this and `continuing` will reduce the indent of the loop quite a bit.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+			for _, d := range strings.Split(value, ",") {
+				if _, _, _, err := ParseQualifiedName(d); err != nil {
```
elezar, line 70: Is this error useful to the user? Would using `IsQualifiedName` be sufficient?

pkg/cdi/container-edits.go
klihub: The called function does nothing if all the mandatory information is already present.

pkg/cdi/container-edits.go
klihub: No. IMO, it does not make sense to omit the minor number. So you either omit both or none. Even if we tried, with the current struct we can't know if a 0 minor number is explicitly set to 0 (which is valid) or omitted, because the structure has the member as `Minor int`, not `Minor *int`.

pkg/cdi/container-edits.go
klihub: No, that's just blindly copied into the returned structure. (IIUC) `rwm` happens to be the fallback default elsewhere, so I went with that. Shouldn't affect the outcome.

pkg/cdi/container-edits.go
klihub: No, it's deliberate choice. Appending nothing to a nil edit returns/keeps the edit nil, instead of converting it to an empty edit.

pkg/cdi/container-edits.go
klihub: Well, this also was a conscious choice, mostly due to where/how we use it (we append during injection, then update the OCI Spec and then the whole thing is forgotten about and garbage collected) and since we don't mutate these in place once they are loaded from a Spec file. Still, it is a valid question if this is the right choice.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	return keys, devices, nil
+}
+
+// AnnotationKey returns a unique annotation key for a K8s device plugin.
+// plugin should be in the format of "vendor.device-type". deviceID is the
+// deviceID passed to the device plugin for the allocation.
+func AnnotationKey(plugin, deviceID string) (string, error) {
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+			keys = append(keys, key)
+		}
+	}
+
+	return keys, devices, nil
+}
+
+// AnnotationKey returns a unique annotation key for a K8s device plugin.
+// plugin should be in the format of "vendor.device-type". deviceID is the
+// deviceID passed to the device plugin for the allocation.
```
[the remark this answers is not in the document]
klihub: Updated as "deviceID is the ID of the device the plugin is allocating" (because that is what we propose
to be used to ensure uniqueness).

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+const (
+	// AnnotationPrefix is the prefix for CDI container annotation keys.
+	AnnotationPrefix = "cdi.k8s.io/"
+)
+
+// AnnotateInjection updates annotations with a plugin-specific CDI device
+// injection request for the given devices. Upon any error a non-nil error
+// is returned and annotations are left intact. By convention plugin should
+// be in the format of "vendor.device-type".
+func AnnotateInjection(annotations map[string]string, plugin string, deviceID string, devices []string) (map[string]string, error) {
```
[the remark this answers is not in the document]
klihub: I understand where you come from. I chose the name from the CDI point of view, not from the (generic) annotations point of view which would have probably led me to choose the name you suggest. IMO, we are annotating the CDI (device) injection here. Normally that will end up looking like `cdi.AnnotateInjection(..., devices)` in the code, which feels very natural to me. So I'd rather leave it like that.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	// AnnotationPrefix is the prefix for CDI container annotation keys.
+	AnnotationPrefix = "cdi.k8s.io/"
```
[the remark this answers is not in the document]
klihub: Not in my opinion. We don't want to consider things like "cdi.k8s.ioumoney" a CDI annotation and having the prefix defined to include the slash is probably the easiest way to ensure that without extra code.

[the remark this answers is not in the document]
elezar: Yes, that's also true. Thanks for the clarification.

pkg/cdi/annotations.go
klihub: Updated accordingly.

pkg/cdi/annotations.go
klihub: You're right, probably only useful when dealing with specs, not when annotations. Updated to use `IsQualifiedName` and an error message without the specific details.

pkg/cdi/annotations.go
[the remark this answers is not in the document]
elezar: I don't agree that we are annotating the injection. The annotations are just a mechanism for defining the devices that need to be injected. From my perspective, if this were a method, the receiver would be the `annotations` and as such the verb should indicate the operation being performed on them.

Maybe argued from a different angle. This function and the `InjectDevices` function are used in two completely separate components. The first (e.g. a device plugin) may not be aware of what "inject" means. It is merely updating the annotations to ensure that the runtime has the required information.

What about `UpdateAnnotations` instead?
```
annotationsWithCDI, err := cdi.UpdateAnnotations(annotations, plugin, deviceID, devices)
annotations = annotationsWithCDI
```

Also, are maps passed by value, or would the current implementation modify the existing `annotations` making the return value redundant?

pkg/cdi/container-edits.go
elezar: For clarification, I'm OK with not solving this now.

klihub: Okay. Let's sit on this a bit and if it feels like (deep-)copying (device nodes) would be a better choice, we can update that with a separate PR.

pkg/cdi/annotations.go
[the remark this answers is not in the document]
klihub: Okay. How would `AnnotateDevices()` sound to you ?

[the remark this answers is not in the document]
klihub: Uhmmkay... I bit the bullet and changed it to `UpdateAnnotations()`.

pkg/cdi/annotations.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return keys, devices, nil
+}
+
+// AnnotationKey returns a unique annotation key for a K8s device plugin.
+// plugin should be in the format of "vendor.device-type". deviceID is the
+// deviceID passed to the device plugin for the allocation.
+func AnnotationKey(plugin, deviceID string) (string, error) {
+	const maxNameLen = 63
```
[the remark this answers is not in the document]
kad: IIRC, it comes from k8s annotation/label name limits.

elezar (approved):
Thanks @klihub. Sorry for the delay.

klihub:
> Thanks @klihub. Sorry for the delay.

No problemo, I think this was pretty fast. Thanks for the review !

PR #51, opened 2022-03-10
update dependencies to recent versions

bart0sh:


go.mod
```diff
@@ -4,11 +4,11 @@ go 1.17
 
 require (
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/opencontainers/runc v1.0.3
+	github.com/opencontainers/runc v1.1.0
```
elezar, line 7: Should we update the spec too?

go.mod
```diff
@@ -18,17 +18,16 @@ require (
 require (
 	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/fsnotify/fsnotify v1.5.1 // indirect
 	github.com/hashicorp/errwrap v1.0.0 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
-	github.com/opencontainers/selinux v1.9.1 // indirect
+	github.com/opencontainers/selinux v1.10.0 // indirect
```
elezar, line 23: Was this not changed by @klihub explicitly in order to get the cri-o CI to pass?

klihub: Yes. @elezar is right. We must hold back such a change until this is done on the CRI-O side and the related CRI-O kubelet/OpenShift CI failures are solved.

klihub (changes requested):
As @elezar correctly pointed out, we can't update selinux to that version until the related problems (unexplained OpenShift CI kubernetes cluster bootstrapping failures) are solved and this same change is done first in CRI-O.

elezar:
> As @elezar correctly pointed out, we can't update selinux to that version until the related problems (unexplained OpenShift CI kubernetes cluster bootstrapping failures) are solved and this same change is done first in CRI-O.

We could update it, but that would mean that we may run into similar problems when updating the dependency in the crio codebase again.

go.mod
bart0sh: can you elaborate a bit more?

go.mod
bart0sh: thank you for the review. reverted.

However, go mod tidy changes dependencies this way:
```
diff --git a/go.mod b/go.mod
index b2cd25e..7974104 100644
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
        github.com/davecgh/go-spew v1.1.1 // indirect
        github.com/hashicorp/errwrap v1.0.0 // indirect
        github.com/inconshreveable/mousetrap v1.0.0 // indirect
-       github.com/opencontainers/selinux v1.9.1 // indirect
+       github.com/opencontainers/selinux v1.10.0 // indirect
        github.com/pmezard/go-difflib v1.0.0 // indirect
        github.com/sirupsen/logrus v1.8.1 // indirect
        github.com/spf13/pflag v1.0.5 // indirect
diff --git a/go.sum b/go.sum
index df2375a..fc217e3 100644
--- a/go.sum
+++ b/go.sum
@@ -295,9 +295,6 @@ github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417 h1:3
 github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
 github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e h1:2Tg49TNXSTIsX8AAtmo1aQ1IbfnoUFzkOp7p2iWygtc=
 github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e/go.mod h1:r3f7wjNzSs2extwzU3Y+6pKfobzPh+kKFJ3ofN+3nfs=
-github.com/opencontainers/selinux v1.8.2/go.mod h1:MUIHuUEvKB1wtJjQdOyYRgOnLD2xAPP8dBsCoU0KuF8=
-github.com/opencontainers/selinux v1.9.1 h1:b4VPEF3O5JLZgdTDBmGepaaIbAo0GqoF6EBRq5f/g3Y=
-github.com/opencontainers/selinux v1.9.1/go.mod h1:2i0OySw99QjzBBQByd1Gr9gSjvuho1lHsJxIJ3gGbJI=
 github.com/opencontainers/selinux v1.10.0 h1:rAiKF8hTcgLI3w0DHm6i0ylVVcOrlgR1kK99DRLDhyU=
 github.com/opencontainers/selinux v1.10.0/go.mod h1:2i0OySw99QjzBBQByd1Gr9gSjvuho1lHsJxIJ3gGbJI=
 github.com/pascaldekloe/goe v0.0.0-20180627143212-57f6aae5913c/go.mod h1:lzWF7FIEvWOWxwDKqyGYQf6ZUaNfKdP144TG7ZOy1lc=
```

We should explicitly freeze them I guess. Not sure it's possible though.

go.mod
elezar: Sorry, I thought that the the spec may also be updated, but there is still only a [`v1.02` tag](https://github.com/opencontainers/runtime-spec/releases/tag/v1.0.2).

elezar:
As soon as @klihub is happy with this, I'm good to merge.

elezar (approved):

bart0sh:
@klihub I've updated this PR according to your suggestions. Please review again and merge if you're ok with it.

klihub:
I can't get this working. Unless I revert the runc update back to 1.0.3, go insists updating selinux to 1.10.0.
@bart0sh @elezar Doesn't this PR exhibit the same behavior for you guys ?

bart0sh:
@klihub yep, I've mentioned it in [this comment](https://github.com/container-orchestrated-devices/container-device-interface/pull/51#discussion_r824529417)

We should probably drop this PR. At least for now.

go.mod
```diff
@@ -6,9 +6,9 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/opencontainers/runc v1.0.3
 	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
-	github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e
+	github.com/opencontainers/runtime-tools v0.9.0
```
klihub, line 9: This change will break builds. We use `AddMultiplePorcessEnv()` from runtime-tools which is not part of any tagged release, so it requires a SHA1-exact dependency.

klihub: Of course, if this is a show-stopper then we can rework CDI to use a much older version and later switch back to using `AddMultipleProcessEnv`, if that ever gets covered by a new tag.

klihub: > Not sure it will be ever covered by a new tag.

Hmm, now I'm not sure what you mean. Do you mean that it is possible that there will never be any newer tagged version of `runtime-tools` than the current `0.9.0` ? Wouldn't that sort of imply that there would never be any future updates to the `OCI Spec/runtime-spec package` ?

PR #53, opened 2022-03-14
pkg/cdi: handle unmarshalling empty Spec data.

klihub:
Unmarshalling empty Spec data results in the raw Spec pointer set to nil. Handle this gracefully instead of crashing on it by recording a Spec-specific error indicating the lack of valid Spec data.

### DIFF
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -68,6 +68,9 @@ func ReadSpec(path string, priority int) (*Spec, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to parse CDI Spec %q", path)
 	}
+	if raw == nil {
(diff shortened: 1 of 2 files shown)


elezar (commented):
What is the benefit of not returning a pointer to an empty spec instead? Especially since we don't validate the spec at present?

bart0sh:
> What is the benefit of not returning a pointer to an empty spec instead?

Empty spec is useless I guess and it's better to consider it as an error in my opinion.

elezar:
> > What is the benefit of not returning a pointer to an empty spec instead?
> 
> Empty spec is useless I guess and it's better to consider it as an error in my opinion.

Yes, I agree. What happens when we try to parse `"{}"` (an empty JSON document)? as opposed to `""` a completely empty document?

elezar (approved):
Looks good. Thanks for the discussion.

PR #54, opened 2022-03-16
devel: perform schema-based Spec validation while loading Specs.

klihub:
This PR implements automatic schema-based validation while populating the cache/loading CDI Spec files.


schema/schema.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed remarshal as JSON")
+		}
+	}
+
+	return s.validate(schema.NewStringLoader(string(data)))
+}
+
+// ValidateFile validates the given JSON file against the schema.
+func (s *Schema) ValidateFile(path string) (error, error) {
```
elezar, line 148: It might be useful to add a `Validate` method that accepts an `io.Reader` which is simpler to create from both paths and byte slices.

pkg/cdi/doc.go
```diff
@@ -127,4 +127,23 @@
...
+// the filesystem at /etc/cdi/schema/schema.json it is loaded and used
+// as the default validation schema. If such a file is not found or
+// fails to load, an embedded copy of the in repo schema/schema.json
+// file is used.
+//
+// The used validation schema can also be changed programmatically using
+// the SetSchema API convenience function. This functions also accepts
+// the special "builtin" (BuiltinSchemaName) and "none", (NoneSchemaName)
+// schema names to switch to using the builtin embedded schema, or a NOP
+// scheme.
```
elezar, line 148: Nit: typo?

pkg/cdi/schema.go
```diff
@@ -0,0 +1,48 @@
+/*
+   Copyright © 2021 The CDI Authors
```
elezar, line 2: Nit: Should this date be updated or include a range?

schema/schema.go
```diff
@@ -0,0 +1,215 @@
...
+	case source == NoneSchemaName, source == "":
+		return NopSchema(), nil
+	case strings.HasPrefix(source, "file://"):
+	case strings.HasPrefix(source, "http://"):
+	case strings.HasPrefix(source, "https://"):
+	default:
+		if strings.Index(source, "://") < 0 {
+			source, err = filepath.Abs(source)
+			if err != nil {
+				return nil, err
```
elezar, line 110: Do we need more context in this `err` being returned?

elezar (commented):
Thanks @klihub. Looks good. Some minor comments, but I don't think anything is blocking. 

Would like your view on them before I approve / merge though.

pkg/cdi/schema.go
klihub: True, I updated the date.

schema/schema.go
klihub: Yes, I think we do. Thanks for spotting. Updated with a better error message.

schema/schema.go
klihub: Good point. I try to roll such an interface.

klihub: I added such an interface.

klihub:
@elezar, @bart0sh I feel a bit uneasy about enabling schema-based validation by default, mostly due to the fact that we do't really know the [quality/reliability](https://github.com/xeipuuv/gojsonschema/issues) of the [schema validation package](https://github.com/xeipuuv/gojsonschema).

So all things considered, I'm think it would be safer to *disable* runtime validation by default... at least for the case when an external schema is not found at the well-known location (IOW, default to `none` instead of `builtin`). WDYT ?

elezar:
> So all things considered, I'm think it would be safer to _disable_ runtime validation by default... at least for the case when an external schema is not found at the well-known location (IOW, default to `none` instead of `builtin`). WDYT ?

One question, how does a failed schema affect behaviour from a user perspective and what steps should a user follow to bypass a failing validation if it is considered a false positive? If these are well documented and simple workarounds, then I think that we can enable it by default. If we are concerned that these errors will be difficult to deal with -- or have a high number of false positives -- the disabling this by default may be better.

Is there a way for us to collect more data on the stability of the schema validation package?

bart0sh:
> @elezar, @bart0sh I feel a bit uneasy about enabling schema-based validation by default, mostly due to the fact that we don't really know the [quality/reliability](https://github.com/xeipuuv/gojsonschema/issues) of the [schema validation package](https://github.com/xeipuuv/gojsonschema).
> 
> So all things considered, I'm think it would be safer to _disable_ runtime validation by default... at least for the case when an external schema is not found at the well-known location (IOW, default to `none` instead of `builtin`). WDYT ?

It think it definitely makes sense to disable it by default. At least for some time.

klihub:
> > So all things considered, I'm think it would be safer to _disable_ runtime validation by default... at least for the case when an external schema is not found at the well-known location (IOW, default to `none` instead of `builtin`). WDYT ?
> 
> One question, how does a failed schema affect behaviour from a user perspective and what steps should a user follow to bypass a failing validation if it is considered a false positive? If these are well documented and simple workarounds, then I think that we can enable it by default. If we are concerned that these errors will be difficult to deal with -- or have a high number of false positives -- the disabling this by default may be better.
> 
> Is there a way for us to collect more data on the stability of the schema validation package?

That's a good question. If we fully trusted from the quality perspective the validation package (no crashes, security problems, etc), then one way would be to disable it but still run validation and then record (log) the results. But I think we don't know if the package is safe enough for that, and I'd prefer to err on the side of caution, IOW not dare doing it. Then probably the next best option is to test validation ourselves more rigorously, and to make it easy for people to run validation on their Specs and then report failures as bugs.

klihub:
> > @elezar, @bart0sh I feel a bit uneasy about enabling schema-based validation by default, mostly due to the fact that we don't really know the [quality/reliability](https://github.com/xeipuuv/gojsonschema/issues) of the [schema validation package](https://github.com/xeipuuv/gojsonschema).
> > So all things considered, I'm think it would be safer to _disable_ runtime validation by default... at least for the case when an external schema is not found at the well-known location (IOW, default to `none` instead of `builtin`). WDYT ?
> 
> It think it definitely makes sense to disable it by default. At least for some time.

Okay, since we all seem/tend to agree, I updated the PR to default to `none` schema/disabled validation.

bart0sh (approved):
/lgtm

elezar (approved):
Looks good @klihub. Thanks for raising the question of the default validator.

PR #56, opened 2022-03-21
pkg/cdi: implement automatic CDI registry refresh.

klihub:
Implement automatic refresh of the registry. It is enabled by default and can be disabled using the newly introduced `WithoutSelfRefresh()` option. When in automatic/self-refresh mode, CDI Spec directories are monitored for changes. A `Refresh()` is internally triggered whenever a change to a Spec file or directory is detected.


elezar (commented):
I have not had a look at the changes specifically, but from your pros and cons, I think I would lean towards allowing parallel injections and lighter weight locking over handling directories being created or removed.

I would think that on startup a missing directory could either be treated as an error, or the missing directory is skipped and not stored in the cache. This means that updates to existing directories are refreshed automatically but we need not handle the addition of directories. This may not quite handle directories being removed though.

klihub:
> I have not had a look at the changes specifically, but from your pros and cons, I think I would lean towards allowing parallel injections and lighter weight locking over handling directories being created or removed.
> 
> I would think that on startup a missing directory could either be treated as an error, or the missing directory is skipped and not stored in the cache. This means that updates to existing directories are refreshed automatically but we need not handle the addition of directories. This may not quite handle directories being removed though.

Okay, any other takes on this ? @bart0sh, @kad ? 

My stand on these is roughly the following. My strongest dislike(s) for the naive approach are 1) There is no easy reasonable default reaction to a directory missing on startup because depending on whether you are starting up a runtime client (a pure CDI consumer), a runtime (another type of pure CDI consumer), or for instance a device plugin (both a producer and in a way also an indirect consumer), the reasonable default actions vary between, ignore, create, or fail with an error. 2) What really badly break with the simplistic implementation is removing then recreating a directory. It completely breaks any further tracking of changes without a restart. And a restart is fine for some CDI clients but not for others.

I can try to bring the implementation of the simplistic approach linked to above on par with the one in this PR and see if I can come up with some reasonable remedies for the above problems (apart from the lost tracking bits which is unsolvable IMO).

cmd/cdi/cmd/monitor.go
```diff
@@ -81,49 +81,41 @@ func monitorSpecDirs(args ...string) {
 	done = make(chan error, 1)
 
 	go func() {
+		// don't print the more often than this
```
elezar, line 84: typo: extra `the`

cmd/cdi/cmd/monitor.go
```diff
@@ -81,49 +81,41 @@ func monitorSpecDirs(args ...string) {
...
+		lowPassFilter := 1 * time.Second
```
elezar, line 85: Nit: I think calling this `oneSecond` instead will make the code below more readable.

pkg/cdi/cache.go
```diff
@@ -162,6 +219,11 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	c.Lock()
 	defer c.Unlock()
 
+	// in self-refresh mode trigger a refresh here if a Spec dir appears
+	if c.selfRefresh && c.watch.update() {
```
elezar, line 223: It seems strange to me that this is required here (or only here). Should all queries to the cache not require a refresh?

I would expect the watcher to trigger the refresh on updates to ensure that the cache state is always up to date.

pkg/cdi/cache.go
```diff
@@ -65,17 +84,58 @@ func (c *Cache) Configure(options ...Option) error {
...
+		for _, err := range c.errors {
+			result = multierror.Append(result, err...)
+		}
+		return result
+	}
+
+	// We need to refresh if
+	// - we're in manual refresh mode, or
+	// - in self-refresh mode a missing Spec dir appears (added to watch)
+	if !c.selfRefresh || c.watch.update() {
```
elezar, line 130: Instead of having `selfRefresh` be a member of the cache, would it not make sense to rather instantiate a watcher that always returns `true` for `update()`?

pkg/cdi/cache.go
```diff
@@ -277,3 +339,113 @@ func (c *Cache) GetSpecDirectories() []string {
...
+func (w *watch) update(removed ...string) bool {
+	var (
+		dir    string
+		done   bool
+		err    error
+		update bool
+	)
+
+	for dir, done = range w.watched {
+		if done {
```
elezar, line 394: Does this handle directories being removed from the filesystem? If I read this correctly, once a directory is added set to `watched[dir] = true` there is no way for this to be set to false unless it is removed explicitly.

pkg/cdi/cache.go
```diff
@@ -277,3 +339,113 @@ func (c *Cache) GetSpecDirectories() []string {
...
+			continue
+		}
+		err = w.w.Add(dir)
+		switch {
```
elezar, line 398: Is there a particular reason for a switch statement here instead of an `if-else` or `if-continue`?

pkg/cdi/cache.go
```diff
@@ -277,3 +339,113 @@ func (c *Cache) GetSpecDirectories() []string {
...
+			w.watched[dir] = true
+			update = true
+		case !os.IsNotExist(err):
+			fallthrough
+		default:
+			w.watched[dir] = false
+		}
+	}
+
+	for _, dir = range removed {
```
elezar, line 409: Does it not make sense to remove these first and then update `w.watched`?

pkg/cdi/cache.go
elezar: OK, looking at the code a bit more, the the `removed` args mark a directory as removed and doesn't actually remove it from the list.

pkg/cdi/cache.go
```diff
@@ -277,3 +339,113 @@ func (c *Cache) GetSpecDirectories() []string {
...
+				continue
+			}
+			if event.Op == fsnotify.Write {
+				if ext := filepath.Ext(event.Name); ext != ".json" && ext != ".yaml" {
+					continue
+				}
+			}
+
+			m.Lock()
+			var removed []string
```
elezar, line 437: There is no need to construct this slice since `update` accepts a variadic list.

elezar (commented):
Thanks @klihub I have had a bit more of a chance to look into this and have some comments.

cmd/cdi/cmd/monitor.go
klihub: Fixed.

cmd/cdi/cmd/monitor.go
klihub: Updated.

pkg/cdi/cache.go
klihub: Yes, in principle all of them should have this same snippet to guarantee 100% correct, up-to-date results. This is the only place though which is currently critical from the point of view of runtimes. I can plant it to the rest of the places, too.

pkg/cdi/cache.go
klihub: Directories are removed when they are removed form the list of spec directories (IOW, the registry is re-Configure()'d). Otherwise it is either marked watched (directory exists) or not watched (directory absent) until the directory is recreated.

pkg/cdi/cache.go
klihub: It can be done like that, too. I'm not sure though if that matters at all. I think it does not. The idea here is that we always check here if we could add a pending spec directory, because we don't get any external events for the creation of those. We do get events from subsequent removal, which we then handle, if we were triggered by a removal event.

pkg/cdi/cache.go
klihub: That would make it awkward to keep the current test/semantics in `InjectDevices`, which only `refresh()`es internally if we are in self-refresh mode and the watch got updated (missing dir created). If we removed the explicitly stored `selfRefresh` we'd need to do something like, create `watch.w` when `selfRefresh` is set, `Close()` and set it to `nil` when `selfRefresh` is cleared, and use a `watch.w != nil` test  for approximating `selfRefresh`.

pkg/cdi/cache.go
klihub: That was done that way to set up the arguments in the if-else then call `update()` on the common path... which is not that helpful. Updated pass args without a slice and the ...-operator.

pkg/cdi/cache.go
```diff
@@ -238,6 +315,11 @@ func (c *Cache) ListClasses() []string {
 	c.Lock()
 	defer c.Unlock()
 
+	// in self-refresh mode trigger a refresh here if a Spec dir appears
+	if c.selfRefresh && c.watch.update(c.dirErrors) {
```
elezar, line 319: Question: Since we're using this regularly now, does it make sense to wrapt this in a method `mustRefresh`? or even `refreshIfRequired()` to improve readability and remove the need to keep repeating this comment?

pkg/cdi/registry.go
```diff
@@ -50,10 +52,15 @@ type Registry interface {
...
 // order of Refresh().
+//
+// GetSpecDirErrors returns any errors related to the configured
+// Spec directories.
 type RegistryRefresher interface {
+	Configure(...Option) error
 	Refresh() error
 	GetErrors() map[string][]error
 	GetSpecDirectories() []string
+	GetSpecDirErrors() map[string]error
```
elezar, line 63: I suppose one question would be why these errors are not included in the `GetErrors`?

elezar (commented):
Thanks @klihub. This is actually looking really good now (Probably because I've started understanding it better). I have some minor questions / comments, but I don't think they're blockers.

pkg/cdi/cache.go
```diff
@@ -33,30 +34,46 @@ type Option func(*Cache) error
...
+	specs     map[string][]*Spec
+	devices   map[string]*Device
+	errors    map[string][]error
+	dirErrors map[string]error
+
+	selfRefresh bool
+	watch       *watch
+}
+
+// WithSelfRefresh returns an option to control automatic Cache refresh.
```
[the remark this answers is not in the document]
klihub: Yes, I agree. `AutoRefresh` is a much better and less confusing term. I'll update all related terms like that.

pkg/cdi/cache.go
```diff
@@ -273,7 +360,142 @@ func (c *Cache) GetErrors() map[string][]error {
...
+	return errors
+}
+
+// Our fsnotify helper wrapper.
+type watch struct {
+	w       *fsnotify.Watcher
+	watched map[string]bool
+}
+
+// Set up monitoring for the given Spec directories.
```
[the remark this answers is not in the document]
klihub: Updated comment accordingly.

pkg/cdi/cache.go
```diff
@@ -273,7 +360,142 @@ func (c *Cache) GetErrors() map[string][]error {
...
+// GetSpecDirErrors returns any errors related to configured Spec directories.
+func (c *Cache) GetSpecDirErrors() map[string]error {
+	if c.dirErrors == nil {
+		return nil
+	}
+
+	c.Lock()
+	defer c.Unlock()
+
+	errors := make(map[string]error)
```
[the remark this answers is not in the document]
klihub: Now that you asked, I had to check from the docs. Unfortunately it can't be used. The signature requires slices (although it also works on strings as a source).

pkg/cdi/cache.go
klihub: Good point and also much less error-prone for future changes. Updated accordingly.

pkg/cdi/cache.go
```diff
@@ -273,7 +349,142 @@ func (c *Cache) GetErrors() map[string][]error {
...
+	for dir, err := range c.dirErrors {
+		errors[dir] = err
+	}
+	return errors
+}
+
+// Our fsnotify helper wrapper.
+type watch struct {
+	w       *fsnotify.Watcher
```
bart0sh, line 378: readability nit: rename w -> watcher to avoid using ```w.w``` in the below receiver functions.

pkg/cdi/cache.go
```diff
@@ -33,30 +34,46 @@ type Option func(*Cache) error
...
+		autoRefresh: true,
+		watch:       &watch{},
 	}
 
-	return c, c.Refresh()
+	WithSpecDirs(DefaultSpecDirs...)(c)
+	c.Lock()
+	defer c.Unlock()
+
+	return c, c.configure(options...)
```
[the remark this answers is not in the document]
klihub: I'm not sure it that'd be possible... I think it is not. A similarly straightforward "always take the lock on the ingress" approach was possible and that's how it used to be when `[Rr]efresh()` was always triggered externally. Now that we trigger it internally, both by `inotify` events and depending on the circumstances by CDI clients making calls into the package, we can't have it so simple any more. `refresh()` is now called from very different contexts and it can't assume it can take the lock. So it has to assume the lock has been taken on its behalf by the caller.

pkg/cdi/registry.go
klihub: Good question... the alternative branch also includes these in the errors returned by `GetErrors()` so I updated this accordingly.

They are tracked separately because their semantics/lifecycle is a bit different (watch vs. refresh). The querying function exists to make it easy to dig out the errors related to `Spec` dirs as opposed to `Spec` content. Did you mean that it would be enough to return these among `GetErrors()` ?

elezar: Yes, I was proposing that these be returned along with `GetErrors()`. I understand their lifecycle is different, but I don't know whether that justifies exposing it in the interface.

As an example, if a user were to call `Refresh()` do we expect that they call both `GetErrors()` and `GetSpecDirErrors()` to check the status of the refresh call? In which cases would they call one or the other?

klihub: No, that should not be necessary. `GetErrors()`, reports the same Spec dir errors, too.

It's the other way around I had in mind with this. If a client wants to know (for instance after having updated the configuration) whether there is anything wrong with the configured Spec directory configuration, then `GetSpecDirErrors()` would be the straightforward way to do that.

pkg/cdi/cache.go
klihub: Updated accordingly.

bart0sh (approved):
/lgtm

pkg/cdi/registry.go
elezar: I still don't think we should make the interface unnecessarily broad, but I'm not going to block this PR on that. The use cases you describe are valid.

pkg/cdi/cache.go
```diff
@@ -65,17 +82,60 @@ func (c *Cache) Configure(options ...Option) error {
...
+		for _, err := range c.errors {
+			result = multierror.Append(result, err...)
+		}
+		return result
+	}
+
+	// We need to refresh if
+	// - we're in manual refresh mode, or
+	// - in auto-refresh mode a missing Spec dir appears (added to watch)
+	if !c.autoRefresh || c.watch.update(c.dirErrors) {
```
elezar, line 130: It seems a little strange to have this logic here and then use `refreshIfRequired()` elsewhere. Does it make sense to add a `force bool` argument to `refreshIfRequired` and then to call this here as:
```
refreshIfRequired(!c.autoRefresh)
```
(dropping the conditional).

pkg/cdi/cache.go
```diff
@@ -65,17 +82,60 @@ func (c *Cache) Configure(options ...Option) error {
...
+		return c.refresh()
+	}
+
+	return cachedErrors()
```
elezar, line 134: Does this mean that we lose access to the `cachedErrors` if we end up calling `refresh`? Is this intentional?

pkg/cdi/cache.go
```diff
@@ -149,6 +206,13 @@ func (c *Cache) Refresh() error {
 	return nil
 }
 
+// RefreshIfRequired triggers a refresh with auto-refresh if Spec dirs have appeared.
+func (c *Cache) refreshIfRequired() {
+	if c.autoRefresh && c.watch.update(c.dirErrors) {
```
elezar, line 211: See my earlier comment about possibly adding a `force` argument here.

pkg/cdi/cache.go
```diff
@@ -65,17 +82,60 @@ func (c *Cache) Configure(options ...Option) error {
...
+	// - we're in manual refresh mode, or
+	// - in auto-refresh mode a missing Spec dir appears (added to watch)
+	if !c.autoRefresh || c.watch.update(c.dirErrors) {
+		return c.refresh()
+	}
+
+	return cachedErrors()
+}
+
+// Rescan CDI Spec directories and refresh the Cache.
```
elezar, line 137: Nit: this docstring isn't of the recommended format.

elezar (approved):
Some minor comments on internal working. These could be addressed here or as follow ups.

Once again thanks for the great work @klihub

pkg/cdi/cache.go
klihub: No, it's sort of the opposite... we want to keep refresh idempotent with respect to its return value.

If we don't end up calling `refresh()` in case we're already up-to-date, then we still return the cached errors from the last `refresh()` to keep it idempotent. `cachedErrors()` is simply used as a macro here. It's a locally-scoped function pointer. Since it is only used in one place, if it is considered disturbing it can be replaced easily with the body of the anonymous function and the accompanied comment.

pkg/cdi/cache.go
klihub: Updated.

pkg/cdi/cache.go
klihub: True, it might easier to understand if we distill all the remaining related logic to `refreshIfRequired()`. I'll update it.

pkg/cdi/cache.go
klihub: Rewrote, distilling the former manual `Refresh()` logic here, too.

bart0sh:
@elezar can you review again and merge? It looks like @klihub addressed your comments.

elezar:
> @elezar can you review again and merge? It looks like @klihub addressed your comments.

Yes, he pinged me offline and asked me to hold off on merging to run one more test. I just wanted to be sure that this was completed. As soon as I get his go-ahead I will review this again.

mikebrow:
reviewing..

pkg/cdi/spec-dirs_test.go
```diff
@@ -242,8 +242,12 @@ func updateTestDir(t *testing.T, tmp string, dirs map[string]map[string]string)
 		}
 		for file, data := range content {
 			file := filepath.Join(dir, file)
-			if err := ioutil.WriteFile(file, []byte(data), 0644); err != nil {
-				return errors.Wrapf(err, "failed to write file %q", file)
+			tmp := file + ".tmp"
+			if err := ioutil.WriteFile(tmp, []byte(data), 0644); err != nil {
+				return errors.Wrapf(err, "failed to write file %q", tmp)
+			}
+			if err := os.Rename(tmp, file); err != nil {
```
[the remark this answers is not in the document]
klihub: A subsequent commit at the tip of [this tree](https://github.com/klihub/container-device-interface/commits/devel/spec-write), which implements a simple API for writing Spec files, mirrors this same behavior.

The problem with first testing for the existence of a file and only then creating/writing it is that this is not atomic. There is a time window between the check and the creation operations where another process can create the exact file which was verified not to exist. If we want the API to provide means for preventing the replacement of an existing file we can do that by 
  - letting the caller explicitly indicate whether overwrite is desired (`Spec.Write(replace bool) error`)
  - using `[golang.org/x/sys/]unix.Renameat2(tmp, file, unix.RENAME_NOREPLACE)` for atomic check/replace

I can change the implementation if this is considered useful. @kad @elezar @bart0sh, WDYT ? 

So far this has not been a concern mostly due to the Spec file structure which syntactically limits a single Spec file to declare only instances of a single vendor/device-type combo. This implies a strong vendor-ownership of every Spec file, which then hopefully gives a strong hint for vendors/implementations to namespace their Spec files with a unique vendor-specific prefix.

cmd/cdi/cmd/monitor.go
```diff
@@ -81,49 +81,41 @@ func monitorSpecDirs(args ...string) {
...
 	go func() {
+		// don't print registry content more often than this
+		oneSecond := 1 * time.Second
+
 		if len(args) == 0 {
 			args = []string{"all"}
 		}
 
-		cdiPrintRegistry(args...)
+		timer := time.NewTimer(oneSecond)
```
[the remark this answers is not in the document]
klihub: Updated code with a deferred timer.Stop().

pkg/cdi/spec-dirs_test.go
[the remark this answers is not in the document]
mikebrow: fair.. if the rename fails due to exist.. should we "fail" or resolve

elezar:
@mikebrow since all comments have been resolved, could we get a "lgtm" or approval if applicable?

mikebrow:
> @mikebrow since all comments have been resolved, could we get a "lgtm" or approval if applicable?

was waiting on response from: 

>  If we want the API to provide means for preventing the replacement of an existing file we can do that by
> - letting the caller explicitly indicate whether overwrite is desired (Spec.Write(replace bool) error)
> - using [golang.org/x/sys/]unix.Renameat2(tmp, file, unix.RENAME_NOREPLACE) for atomic check/replace
> I can change the implementation if this is considered useful. @kad @elezar @bart0sh, WDYT ?I can change the implementation if this is considered useful. @kad @elezar @bart0sh, WDYT ?

pkg/cdi/spec-dirs_test.go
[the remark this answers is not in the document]
elezar: I think we should "fail"

Assuming that the caller can determine the desired filename, I would vote for keeping the API implementation simpler in that if a file already exists at the target we would error out (whether this is done atomically or not). The caller can then decide whether overwriting the file is the desired behaviour -- removing the existing file explicitly.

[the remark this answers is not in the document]
elezar: With that said, since this is currently test code we can address these questions in the upcoming API additions for writing spec files. Would that be reasonable from your perspective @mikebrow?

klihub:
> > @mikebrow since all comments have been resolved, could we get a "lgtm" or approval if applicable?
> 
> was waiting on response from:
> 
> > If we want the API to provide means for preventing the replacement of an existing file we can do that by
> > 
> > * letting the caller explicitly indicate whether overwrite is desired (Spec.Write(replace bool) error)
> > * using [golang.org/x/sys/]unix.Renameat2(tmp, file, unix.RENAME_NOREPLACE) for atomic check/replace
> > 
> > I can change the implementation if this is considered useful. @kad @elezar @bart0sh, WDYT ?

So, @bart0sh @kad ? Any preferences, opinions ?

pkg/cdi/spec-dirs_test.go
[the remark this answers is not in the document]
bart0sh: @klihub
+1 for letting the caller explicitly indicate whether overwrite is desired 

not sure if we need to use Renameat2 as some filesystems might not support it
Can we simply use os.OpenFile with O_CREATE when caller doesn't want to overwrite it?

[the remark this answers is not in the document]
klihub: > I think we should "fail"
> 
> Assuming that the caller can determine the desired filename, I would vote for keeping the API implementation simpler in that if a file already exists at the target we would error out (whether this is done atomically or not). The caller can then decide whether overwriting the file is the desired behaviour -- removing the existing file explicitly.

@elezar Okay, so the primary consumer for the Spec Write() interface IMO is a CDI-enabled K8s device plugin, for instance an NVidia GPU device plugin. How would the logic for generating/updating/removing CDI Spec file entries/Spec files for that device plugin would look like wrt. K8s/DP events (device allocation and in the future potential device release notification) ? My assumption/target was that the device plugin would

  - (prefer to) use a single Spec file (clear ownership of spec files vs. device plugins)
  - atomically update the Spec file for every plugin state change (allocation, some kind of release notification) with the new content reflecting the updated state

This was the major reason I wanted to add the Spec Write interface and with the semantics I proposed.

But this might be a total misconception on my side. Therefore I think it would be a good idea if you guys (@elezar @bart0sh ) could think this through to understand whether there are any details in the typical/probable DP-generated Spec file lifecycle that a DP could easily get wrong. If there is, then it would be a good idea to provide a single correct implementation, with the right semantics, in the CDI package, so that DP authors could then use that.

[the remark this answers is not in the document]
elezar: That is a fair point. I wasn't really considering the device plugin use case at this stage. I was thinking from a CLI perspective where I (as a device vendor) want to provide a utility to generate spec files for available devices that can be run from the command line in response to some package installation. In this case I want to be able to provide feedback to the user if something unexpected happens.

With that said, I think having a flag `replace` / `overwrite` that is passed to the `Write` method would be able to handle both cases. Meaning that it would fail if `replace` is not `true`.

May I suggest that we create the PR that you are discussing with the addition of this API and then move this discussion there. Although it is valuable, I think these nuanced discussions are out of scope for this particular test utility. Once we are working with the actual API we may also be able to reason about the changes better and provide usage examples.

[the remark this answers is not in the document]
klihub: Yes, I agree that this definitely belongs to a separate PR. Just wanted to bring up that PoV while we were at it, since/once the topic came up. I can file a Spec Write API PR and we can sort out the details there.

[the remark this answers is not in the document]
klihub: I filed that Spec Write API PR as #64. It is marked as draft because one of its test cases will need a small update once this auto-refresh PR is merged.

elezar:
> > @mikebrow since all comments have been resolved, could we get a "lgtm" or approval if applicable?
> 
> was waiting on response from:
> 
> > If we want the API to provide means for preventing the replacement of an existing file we can do that by
> > 
> > * letting the caller explicitly indicate whether overwrite is desired (Spec.Write(replace bool) error)
> > * using [golang.org/x/sys/]unix.Renameat2(tmp, file, unix.RENAME_NOREPLACE) for atomic check/replace
> > 
> > I can change the implementation if this is considered useful. @kad @elezar @bart0sh, WDYT ?

@mikebrow we have moved the discussion to #64. Could we close this PR out given that the discussion here is around a test utility. @klihub will update this again as part of #64

klihub:
@mikebrow, @elezar Based on our chat on Slack we'll then go ahead and merge this. Then we'll update #64 (with a few test updates related to this PR) and merge that as well. ETA tomorrow afternoon EET, so please anyone chime in before that if there are still some related concerns/comments.

elezar:
> @mikebrow, @elezar Based on our chat on Slack we'll then go ahead and merge this. Then we'll update #64 (with a few test updates related to this PR) and merge that as well. ETA tomorrow afternoon EET, so please anyone chime in before that if there are still some related concerns/comments.

I have merged the PR. We will at least address #64 and #67 before publishing the next tag (`v0.4.1`) so we can address any additional comments / concerns before then.

PR #62, opened 2022-04-07
Don't force downstream code pulling in (the huge) JSON Schema validation package.

### LINKED ISSUE #61: Please do not use init functions in library code
When updating container-device-interface from 0.3.0 to 0.3.1 we see a over 500 KB binary size increase in podman, https://github.com/containers/podman/pull/13610
However podman uses only small parts of the code in pkg/cdi, such size increase does not seem reasonable to me. Podman does not use any new logic.
Unless I am overlooking something the problem seems to be caused by https://github.com/container-orchestrated-devices/container-device-interface/pull/54
The use of init functions has two major problems: 
 - First, it always runs on program startup so this slow each podman command down, even when most code paths will never even call CDI.

- Second, it adds a lot unnecessary binary size because it calls code that way never be used otherwise by the library user. This is made worse by the fact that the PR started to embed the json spec into the binary.

I don't have enough knowledge about how this library is used to suggest a good fix.

klihub:
Avoid unconditionally importing the Spec JSON Schema validation bits.

The external packages we depend on for that (`github.com/xeipuuv/gojson{schema,reference,pointer}`) are huge. For instance in `cmd/cdi` importing them results in a binary size increase of cc. `3000 kB`. We neither want nor can force such a penalty on anybody that imports CDI.

To avoid this, we split out the Schema-validation glue code bits to a `validate` subpackage and introduce an externally settable Spec validation function. Using this new setup, simply importing `pkg/cdi` alone keeps the offending bits and the related size penalty out of the compiled binaries. However, any client that needs validation and can take the related size penalty can still enable automatic validation using a few lines of code. See `cmd/cdi` for an example of how to do this.

This should fix #61. Closing that should eventually allow this [CDI update podman PR](https://github.com/containers/podman/pull/13610) moving forward.


schema/schema.go
```diff
@@ -39,7 +39,7 @@ const (
 	// NoneSchemaName references a disabled/NOP schema for Load()/Set().
 	NoneSchemaName = "none"
 	// DefaultSchemaName is the none schema.
```
elezar, line 41: This comment is out of sync now.

pkg/cdi/spec.go
```diff
@@ -35,6 +35,9 @@ var (
 		"0.2.0": {},
 		"0.3.0": {},
 	}
+
+	// Externally set CDI Spec validation function.
+	specValidator func(*cdi.Spec) error
```
elezar, line 40: Not critical for this change, but we could use an interface instead.

elezar (commented):
Thanks @klihub. Minor issue with a comment being out of sync. Otherwise this looks good.

pkg/cdi/spec.go
klihub: You mean something like:

```
type SpecValidator interface {
    Validate(*cdi.Spec) error
}

```
Sure, we can do that. I'm just curious what would be the benefit to it ? Having state for the validator does not require it. Using a member function or a closure (like in `pkg/cdi/validate`) does that. If the idea is to have a more formal type defined, then I'd rather have a typedef'ed function. Something like this

```
// SpecValidator is a function for...
type SpecValidator func(*cdi.Spec) error
...
// SetSpecValidator specifies a function for...
func SetSpecValidator(v SpecValidator) {
    specValidator = v
}
```

schema/schema.go
klihub: Thanks for spotting. Removed it as it is not used now.

pkg/cdi/spec.go
elezar: It would allow us to assign more complex validators at a later stage without requiring modification to the API. I don't feel too strongly about it though, so feel free to leave it as is for the purpose of this PR.

elezar (approved):
Thanks for dealing with this so swiftly @klihub.

It looks good from my perspective.

pkg/cdi/spec.go
klihub: Okay. Then let's keep it this way for now. The only real user for it right now is anyway just the `pkg/cdi.SetSpecValidator(pkg/cdi/validate.With*Spec()`-boilerplate to enable extra, schema-based validation, so currently it's more of an internal interface which just happens to be split up the way it is to keep part of the external dependencies optional.

PR #63, opened 2022-04-08
schema: fix Mount.Options in schema.

klihub:
Spec mount options should be an array of strings not a string.

### DIFF
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -62,7 +62,7 @@
                     "$ref": "#/definitions/FilePath"
                 },
                 "options": {
-                    "type": "string"
+                    "$ref": "#/definitions/ArrayOfStrings"
                 }
             },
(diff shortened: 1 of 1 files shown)


schema/defs.json
```diff
@@ -62,7 +62,7 @@
                     "$ref": "#/definitions/FilePath"
                 },
                 "options": {
-                    "type": "string"
+                    "$ref": "#/definitions/ArrayOfStrings"
```
elezar, line 65: Did we miss this when we updated the spec?

elezar (approved):

schema/defs.json
klihub: I think so. Have a vague recollection that we spotted this earlier... and I though we've merged a fix already. But apparently not.

PR #64, opened 2022-04-14
pkg/cdi: add functions for writing Spec files.

klihub:
Implement API for writing/updating Spec files.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync"
 
+	cdi "github.com/container-orchestrated-devices/container-device-interface/specs-go"
 	"github.com/fsnotify/fsnotify"
 	"github.com/hashicorp/go-multierror"
 	oci "github.com/opencontainers/runtime-spec/specs-go"
@@ -252,6 +253,39 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	}
 
 	return nil, nil
+}
+
+// WriteSpec writes a Spec file with the given content. Priority is used
+// as an index into the list of Spec directories to pick a directory for
+// the file, adjusting for any under- or overflows. If name has a "json"
+// or "yaml" extension it choses the encoding. Otherwise JSON encoding
+// is used with a "json" extension.
+func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
+	var (
+		specDir string
+		path    string
+		prio    int
+		spec    *Spec
+		err     error
+	)
+
+	if len(c.specDirs) == 0 {
+		return errors.New("no Spec directories to write to")
+	}
+
+	prio = len(c.specDirs) - 1
+	specDir = c.specDirs[prio]
+	path = filepath.Join(specDir, name)
+	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
+		path += ".json"
+	}
+
+	spec, err = NewSpec(raw, path, prio)
+	if err != nil {
+		return err
+	}
+
+	return spec.Write(true)
 }
 
 // GetDevice returns the cached device for the given qualified name.
--- a/pkg/cdi/registry.go
+++ b/pkg/cdi/registry.go
@@ -19,6 +19,7 @@ package cdi
 import (
 	"sync"
 
+	cdi "github.com/container-orchestrated-devices/container-device-interface/specs-go"
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 )
 
@@ -97,11 +98,15 @@ type RegistryDeviceDB interface {
 //
 // GetSpecErrors returns any errors for the Spec encountered during
 // the last cache refresh.
+//
+// WriteSpec writes the Spec with the given content and name to the
+// last Spec directory.
 type RegistrySpecDB interface {
 	ListVendors() []string
 	ListClasses() []string
 	GetVendorSpecs(vendor string) []*Spec
 	GetSpecErrors(*Spec) []error
+	WriteSpec(raw *cdi.Spec, name string) error
 }
 
 type registry struct {
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -17,6 +17,7 @@
 package cdi
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -68,7 +69,7 @@ func ReadSpec(path string, priority int) (*Spec, error) {
 		return nil, errors.Wrapf(err, "failed to read CDI Spec %q", path)
 	}
 
-	raw, err := parseSpec(data)
+	raw, err := ParseSpec(data)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to parse CDI Spec %q", path)
 	}
@@ -107,6 +108,56 @@ func NewSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
 	}
 
 	return spec, nil
+}
+
+// Write the CDI Spec to the file associated with it during instantiation
+// by NewSpec() or ReadSpec().
+func (s *Spec) Write(overwrite bool) error {
+	var (
+		data []byte
+		dir  string
+		tmp  *os.File
+		err  error
+	)
+
+	err = validateSpec(s.Spec)
+	if err != nil {
+		return err
+	}
+
+	if filepath.Ext(s.path) == ".yaml" {
+		data, err = yaml.Marshal(s.Spec)
+	} else {
+		data, err = json.Marshal(s.Spec)
+	}
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal Spec file")
+	}
+
+	dir = filepath.Dir(s.path)
+	err = os.MkdirAll(dir, 0o755)
+	if err != nil {
+		return errors.Wrap(err, "failed to create Spec dir")
+	}
+
+	tmp, err = os.CreateTemp(dir, "spec.*.tmp")
+	if err != nil {
+		return errors.Wrap(err, "failed to create Spec file")
+	}
+	_, err = tmp.Write(data)
+	tmp.Close()
+	if err != nil {
+		return errors.Wrap(err, "failed to write Spec file")
+	}
+
+	err = renameIn(dir, filepath.Base(tmp.Name()), filepath.Base(s.path), overwrite)
+
+	if err != nil {
+		os.Remove(tmp.Name())
+		err = errors.Wrap(err, "failed to write Spec file")
+	}
+
+	return err
 }
 
 // GetVendor returns the vendor of this Spec.
@@ -183,8 +234,8 @@ func validateVersion(version string) error {
 	return nil
 }
 
-// Parse raw CDI Spec file data.
-func parseSpec(data []byte) (*cdi.Spec, error) {
+// ParseSpec parses CDI Spec data into a raw CDI Spec.
+func ParseSpec(data []byte) (*cdi.Spec, error) {
 	var raw *cdi.Spec
 	err := yaml.UnmarshalStrict(data, &raw)
 	if err != nil {
--- a/pkg/cdi/spec_other.go
+++ b/pkg/cdi/spec_other.go
@@ -1,0 +1,39 @@
+//go:build !linux
+// +build !linux
+
+/*
+   Copyright © 2022 The CDI Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cdi
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// Rename src to dst, both relative to the directory dir. If dst already exists
+// refuse renaming with an error unless overwrite is explicitly asked for.
+func renameIn(dir, src, dst string, overwrite bool) error {
+	src = filepath.Join(dir, src)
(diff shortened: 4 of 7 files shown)


pkg/cdi/cache.go
```diff
@@ -190,6 +191,41 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+		prio    int
+		spec    *Spec
+		err     error
+	)
+
+	if len(c.specDirs) == 0 {
+		return errors.New("no Spec directories to write to")
+	}
+
+	prio = len(c.specDirs) - 1
```
elezar, line 212: This logic seems very difficult to control. Do we have a use case where the caller doesn't just provide the full path?

pkg/cdi/cache.go
```diff
@@ -190,6 +191,41 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+	specDir = c.specDirs[prio]
+
+	if filepath.Ext(name) != "" {
+		path = filepath.Join(specDir, name)
```
elezar, line 216: Should we check for valid extensions (`.yaml`, `.json`).

pkg/cdi/cache.go
```diff
@@ -190,6 +191,41 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+	)
+
+	if len(c.specDirs) == 0 {
+		return errors.New("no Spec directories to write to")
+	}
+
+	prio = len(c.specDirs) - 1
+	specDir = c.specDirs[prio]
+
+	if filepath.Ext(name) != "" {
```
elezar, line 215: This won't behave as we want with dots in the filename. What about something like:
```
path = filepath.Join(specDir, name)
if ext := filepath.Ext(path); ext != "json" && ext != "yaml" {
  path = path+".json"
} 
```

pkg/cdi/registry.go
```diff
@@ -90,11 +91,15 @@ type RegistryDeviceDB interface {
...
 // the last cache refresh.
+//
+// WriteSpec writes the Spec with the given content and name to the
+// last Spec directory.
 type RegistrySpecDB interface {
 	ListVendors() []string
 	ListClasses() []string
 	GetVendorSpecs(vendor string) []*Spec
 	GetSpecErrors(*Spec) []error
+	WriteSpec(raw *cdi.Spec, name string) error
```
elezar, line 102: Question: Should the writing of the spec be part of a different interface?

pkg/cdi/spec_unix.go
```diff
@@ -0,0 +1,51 @@
+//go:build aix || darwin || dragonfly || freebsd || linux || netbsd || openbsd || solaris || zos
+// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris zos
+
+/*
+   Copyright © 2021 The CDI Authors
```
elezar, line 5: Nit: The date.

pkg/cdi/spec_other.go
```diff
@@ -0,0 +1,47 @@
...
+func rename(dir, src, dst string, overwrite bool) error {
+	src = filepath.Join(dir, src)
+	dst = filepath.Join(dir, dst)
+
+	_, err := os.Stat(dst)
+	if err == nil && !overwrite {
+		return os.ErrNotExist
+	}
+
+	return Rename(src, dst)
```
elezar, line 46: Where is `Rename` implemented?

pkg/cdi/spec_unix.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+// Rename src in dir to dst in the same directory. If dst already exists
+// refuse renaming with an error unless overwrite is explicitly asked for.
+func rename(dir, src, dst string, overwrite bool) error {
```
elezar, line 31: Question: Is it strictly necessary that these be in the same directory? If so, should this be `renameIn` or something?

pkg/cdi/spec.go
```diff
@@ -108,6 +109,56 @@ func NewSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
...
+	_, err = tmp.Write(data)
+	tmp.Close()
+	if err != nil {
+		return errors.Wrap(err, "failed to write Spec file")
+	}
+
+	err = rename(dir, filepath.Base(tmp.Name()), filepath.Base(s.path), overwrite)
+
+	if err != nil {
+		os.Remove(tmp.Name())
```
bart0sh, line 155: should we try to remove tmp file in any case?

pkg/cdi/spec.go
```diff
@@ -108,6 +109,56 @@ func NewSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
...
+		err = errors.Wrap(err, "failed to write Spec file")
```
bart0sh, line 156: Can we add s.path to the error message?

pkg/cdi/spec_unix.go
klihub: Thanks. Fixed.

pkg/cdi/spec_other.go
klihub: Thanks ! That should have been `os.Rename(src, dst)`.

pkg/cdi/spec_unix.go
klihub: Not in the native syscall (`renameat2()`). It's just the way how we use it as an unexported/internal function. Maybe we should just call it `renameat()`. That's how the syscall would be called in an ideal world if the API would have been gotten right at the first try.

pkg/cdi/cache.go
klihub: Well, the idea is that in the preferable default case the client does not need to know/care/try to control where the Spec file is spit out. This interface is for that. It always places the Spec file into the last configured Spec directory (IOW produces a file with the highest precedence) and overwrites any existing Spec file with the same name.

In case the client needs/wants to know/control, the lower-level `Spec.Write()` API is for that (which this interface also ends up using in the very end).

pkg/cdi/spec.go
klihub: But why ? After a successful `rename()`, there is no tmp file to remove.

pkg/cdi/cache.go
elezar: OK. That make sense. I would update the doc comment to make this clearer though since the priority cannot be set by the user.

pkg/cdi/cache.go
klihub: Updated accordingly.

pkg/cdi/spec.go
bart0sh: true. overlooked it, sorry.

pkg/cdi/spec_unix.go
bart0sh: I'd suggest to call it renameAt2 as renameAt could make people think that it uses renameat syscall.

pkg/cdi/registry.go
klihub: I deliberately tried to avoid introducing a new one, because TBH I'm not a keen fan of the existing ones either. But if folks consider it better to add a one for this, then please let me know and I'll update the PR accordingly. @bart0sh , @kad ?

bart0sh: @klihub can you elaborate a little bit on this? What should and what shouldn't be in this interface in your opinion and why?

klihub: @bart0sh Sorry, I was too vague in my comment. My comment/dislike was not about the content of the current registry-specific interfaces. It was about the mere existence of those interfaces.

In my original implementation there were no registry-specific extra interfaces, as IIRC there wasn't even a separate `Registry` type at all. The registry was simply a singleton/default instance of the cache and it's full interface was exposed/exported as identically named package-level functions, AFAICT a fairly common golang idiom, a bit similar to how for instance `flag.{Foo(), Xyzzy() and Bar()}` simply alias `flag.CommandLine.{Foo(), Xyzzy(), and Bar()}`.

There was a bit of back and forth and eventually a request to split/introduce the `Registry` interfaces we now have. I agreed to do it so that we could proceed, rework the API to a more usable/consistent set, and get CDI support merged to CRI-O and containerd without much further detail. But, being my admittedly stubborn self, I disliked the idea of the registry being something else than the default instance of the cache already then, and I still dislike it today. Mostly because I don't see any benefit of the extra interfaces, them only causing IMO exactly this kind of unnecessary pain when we extend functionality: which interface should be used in `Registry` to host the new functionality. That's why I tried to avoid introducing one more... 

Does this explain/clarify my earlier comment ?

bart0sh: Yes, it does. I tend to agree with you here. I prefer simplicity over complexity, so if not having additional interfaces  simplifies user experience I'm for it. @elezar wdyt?

elezar: I am happy to remove the interfaces if they're not adding value. It seems as if these were added for a generalisation that was never realised and since the Cache is the only implementation of a registry that we care about (at least at this stage) removing this layer of indirection should not be too problematic.

As a matter of interest, how would returning a cache directly affect our current clients? (`podman` only calls `InjectDevices`).

klihub: The required changes would probably be very minor. My suggestion would be that if we consider removing the extra interfaces, let's cram the Spec Write interface to the currently suggested interface in this PR. Then take a separate look at removing the interfaces, what kind of exact changes it would require in the affected downstream components then, if we decide to go ahead with it, do that in a separate PR.

elezar: That sounds good. Ignore my question of whether this is the right interface for this then and consider this thread resolved.

klihub: Ok, thanks !

pkg/cdi/spec.go
```diff
@@ -109,6 +110,56 @@ func NewSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create Spec file")
+	}
+	_, err = tmp.Write(data)
+	tmp.Close()
+	if err != nil {
+		return errors.Wrap(err, "failed to write Spec file")
+	}
+
+	err = renameAt(dir, filepath.Base(tmp.Name()), filepath.Base(s.path), overwrite)
```
elezar, line 153: Looking at this again: I have two comments. 
1. The arguments of the function `renameAt` here doesn't match those for the `RenameAt` or `RenameAt2` calls used to do the actual rename
2. Using the dir as an argument seems to be more cumbersome than just using filenames. (we can still extract the directories as required in the `rename` implementation).

pkg/cdi/spec.go
```diff
@@ -109,6 +110,56 @@ func NewSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
 	return spec, nil
 }
 
+// Write the CDI Spec to the file associated with it during instantiation
+// by NewSpec() or ReadSpec().
+func (s *Spec) Write(overwrite bool) error {
+	var (
+		data []byte
+		dir  string
+		tmp  *os.File
```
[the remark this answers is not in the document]
klihub: Well, TBH style-wise the `:=`-operator is still more or less a struggle (or at least an ongoing learning experience) for me in golang. I'm not sure whether to like it or hate it. Lately I've tried to pick up a style where I predeclare important variables (algorithmically crucial ones, or ones whose liveness spans a significant part of a function's basic blocks), trying to make the code more readable. But... it's a very subjective thing and I often end up second-guessing myself. I've been reading some of the standard library code in hope of enlightenment... but I'm not sure if one can really get enlightened in this subject. It's just a syntactic ambiguity, with restrictions and too many subjective pros and cons in each case.

pkg/cdi/spec.go
klihub: The current function signature is not an accident. We do want to enforce the two files to already be in the same directory to make sure they are on the same mounted device/filesystem. Otherwise `renameat2()` would fail. If the differing signature wrt. renameat{,2} is too disturbing, I'd suggest we call it `ReplaceAt()` instead, with the same signature as now.

pkg/cdi/spec_other.go
```diff
@@ -0,0 +1,47 @@
...
+// +build !dragonfly
+// +build !freebsd
+// +build !linux
+// +build !netbsd
+// +build !openbsd
+// +build !solaris
+// +build !zos
+
+/*
+   Copyright © 2021 The CDI Authors
```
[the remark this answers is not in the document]
klihub: Thanks! Updated.

pkg/cdi/spec.go
elezar: OK. The check for the same device / filesystem is important in this case. I think `replaceAt` implies an overwrite, so what about `renameIn` instead? From my point of view this makes it clear that the operation is happening in a specified directory with the two files passed as arguments.

I'm also ok to leave it as is if you're strongly against `renameIn` as an option.

pkg/cdi/spec.go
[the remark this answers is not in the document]
elezar: OK. That's fair. I personally prefer the `:=` semantics as I don't generally have to worry about getting the types right (which is especially useful for more complex types) and it generally results in shorter functions and less boiler plate.

pkg/cdi/spec.go
klihub: You are right. `renameIn()` is better, because replacing is behind a boolean flag. I'll update accordingly.

elezar (approved):
Thanks @klihub. This looks good now.

elezar:
@bart0sh @mikebrow any additional comments?

bart0sh (approved):
/lgtm

pkg/cdi/spec_other.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Rename src to dst, both relative to the directory dir. If dst already exists
+// refuse renaming with an error unless overwrite is explicitly asked for.
+func renameIn(dir, src, dst string, overwrite bool) error {
+	src = filepath.Join(dir, src)
+	dst = filepath.Join(dir, dst)
+
+	_, err := os.Stat(dst)
+	if err == nil && !overwrite {
```
klihub, line 42: This is totally hosed. Lemme fix it.

elezar (approved):
Good catch on the overwrite logic.

PR #65, opened 2022-04-19
[spec] Add type field for mounts

### LINKED ISSUE #45: Add a Type field to the Mount spec
The OCI runtime spec's `Mount` struct has a `Type` field that is not currently supported by CDI.

This is used in [runc](https://github.com/opencontainers/runc/blob/eddf35e5462e2a9f24d8279874a84cfc8b8453c2/libcontainer/specconv/spec_linux.go#L508-L510), and without it being set the mount type defaults to a [`proc` mount ](https://github.com/opencontainers/runc/blob/main/libcontainer/rootfs_linux.go#L464-L471).

This leads to errors such as the following when trying to process a mount defined in a CDI `ContainerEdits` (if no options are explicitly specified):
```
Error: OCI runtime error: runc: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "/usr/bin/nvidia-smi" to rootfs at "/usr/bin/nvidia-smi" caused: mount through procfd: no such device
```

Note that if the options includes the `MS_BIND` mount flag, then the [`Type` field is ignored](https://github.com/opencontainers/runc/blob/eddf35e5462e2a9f24d8279874a84cfc8b8453c2/libcontainer/specconv/spec_linux.go#L511-L519).

elezar:
This change adds a Type field for mounts that allows for the specification of, for example, `tmpfs` mounts.

See https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-mounts

Closes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ $ cat > /etc/cdi/vendor.json <<EOF
     ],
     "mounts": [
       {"hostPath": "/bin/vendorBin", "containerPath": "/bin/vendorBin"},
-      {"hostPath": "/usr/lib/libVendor.so.0", "containerPath": "/usr/lib/libVendor.so.0"}
+      {"hostPath": "/usr/lib/libVendor.so.0", "containerPath": "/usr/lib/libVendor.so.0"},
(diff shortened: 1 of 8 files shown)


bart0sh (approved):
/lgtm

klihub (approved):
LGTM.

PR #67, opened 2022-04-29
documentation: add configuration instructions

bart0sh:


klihub (commented):
I'll add my parts...

README.md
```diff
@@ -20,6 +20,22 @@ However, as devices and software grows more complex, vendors want to perform mor
...
+### CRI-O configuration
+
+### Containerd configuration
+
+To enable and configure CDI support in the [containerd runtime](https://github.com/containerd/containerd) 2 configuration options `enable_cdi` and `cdi_spec_dirs` should be set in the `plugins."io.containerd.grpc.v1.cri` section of the containerd configuration file (`/etc/containerd/config.toml` by default):
+
+```
+[plugins."io.containerd.grpc.v1.cri"]
+  enable_cdi = true
+  cdi_spec_dirs = ["/etc/cdi", "/var/run/cdi"]
```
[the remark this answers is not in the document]
elezar: Does the API not already handle the case where this list is empty / unset and use the defaults then?

README.md
```diff
@@ -20,6 +20,22 @@ However, as devices and software grows more complex, vendors want to perform mor
...
+```
+
+Remember to restart containerd for any configuration changes to take effect.
+
```
elezar, line 38: Should we call out how the devices are actually specified?

bart0sh: Can you elaborate? We have a spec format described [here](https://github.com/container-orchestrated-devices/container-device-interface/blob/master/SPEC.md). Devices should be specified according to the spec.

elezar: I mean that for `podman` and as mentioned in the readme below we indicate that the devices are passed through the `--device` command line argument. This is not as clear for containerd as this relies on container annotations, for example. If we don't think that that is relevant, then I'm happy to leave this out.

bart0sh: Thanks. I'll try to come up with something on this regard.

bart0sh: It should be a separate section describing this as the way devices are passed to the CRI runtime are common for CRI-O and Containerd.

klihub: I think from the user experience point of view it probably would be a good idea to give at least one example, as much self-contained as possible, with step-by-step instructions of how to

- set up CDI for the host
  - enable CDI in a/the runtime
  - create/drop CDI configuration in place
- create a container with an injected CDI device
  - using podman
  - using crictl
- examine/verify the result

README.md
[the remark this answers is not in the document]
klihub: And I actually take back this comment of mine. In `cri-o` this is also done similary... so, that in the default configuration `cdi_spec_dirs` is initialized using the exported defaults from `CDI`. Then if you ask `cri-o` to spit out an initial config (`crio config --default`), it simply outputs a configuration file with those defaults already in place. No skulduggery is taking place.

elezar:
> Are these bits still true from the [README](https://github.com/container-orchestrated-devices/container-device-interface/blob/master/README.md#Examples):
> 
> ```
> # Examples
> ...
> # Less verbose, through infering the vendor from the device name
> $ docker/podman run --device myDevice ...
> 
> # Special case
> $ docker/podman run --device vendor.com/device=all ...
> ```
> 
> Unless that logic is implemented on the client side, it does not work like that.

No, those need to be updated. I can update in my PR to this branch as well.

elezar:
@bart0sh what is still required for this PR to be considered complete?

bart0sh:
@elezar is any update for podman instructions required? if so, please update them.

bart0sh:
we still need a step by step instructions of setting up and using CDI as @klihub suggested

README.md
bart0sh: first step 'set up CDI for the host' is more or less covered by this PR. we still need some explanations on how to 'create/drop CDI configuration in place' may be.

remaining 2 parts 'create a container with an injected CDI device' and 'examine/verify the result' need to be added. I'd propose to do it as a real-life example and put into a separate section in the README.md.

elezar:
> we still need a step by step instructions of setting up and using CDI as @klihub suggested

Would it make sense to just use the `example` spec such as:
```
$ cat /etc/cdi/vendor.json | jq
{
  "cdiVersion": "0.2.0",
  "kind": "vendor.com/device",
  "devices": [
    {
      "name": "example",
      "containerEdits": {
        "deviceNodes": [
          {
            "path": "/dev/tty8",
            "type": "c"
          }
        ]
      }
    }
  ],
  "containerEdits": {
    "env": [
      "envVar2=envVal2"
    ]
  }
}
```
for the purpose of this documentation. We are working on tooling to generate specs for NVIDIA Devices and I can update the documentation with links once this is available.

elezar:
> It's ok to use that example. However, it would be better to use real-life example with GPU or FPGA device if that's possible.

The one advantage of the generic example is that any user can run the end-to-end example -- regardless of whether they have access to GPUs or FPGAs. I think this is a starting point and we would expand this with a section on use with GPUs and FPGAs once those are ready.

bart0sh:
Makes sense to me. Let's do it this way.

PR #70, opened 2022-07-04
Relax device name restriction

elezar:
These changes allow device names to start with a digit. This means that fully-qualified device names such as:
```
vendor.com/class=0
```
is now a valid name.


klihub (approved):
LGTM

PR #74, opened 2022-07-12
Don't allow update to empty specDirs

elezar:
This change ensures that the spec dirs for a cache remain
unchanged if the list of directories passed to WithSpecDirs
is empty. This avoids inadvertently disabling CDI with an
empty config.

### DIFF
--- a/pkg/cdi/spec-dirs.go
+++ b/pkg/cdi/spec-dirs.go
@@ -45,6 +45,10 @@ var (
 // WithSpecDirs returns an option to override the CDI Spec directories.
 func WithSpecDirs(dirs ...string) Option {
 	return func(c *Cache) error {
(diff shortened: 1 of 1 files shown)


bart0sh:
@elezar This looks quite implicit to me. We should either trust a caller or consider empty list as an error in my opinion.

bart0sh:
@elezar 
> How does `containerd` handle parsing and forwarding the config options? If the user enables `cdi` but does not specify the spec dirs, are the defaults used?

Yes, defaults are used [when config object is created](https://github.com/containerd/containerd/blob/main/pkg/cri/config/config_unix.go#L109:L110):
They're overwritten only if user specifies options in the config file.

elezar:
OK. Thanks for the input @bart0sh. Let's leave this as is then.

PR #77, opened 2022-08-09
pkg/cdi: add functions for generating Spec names and removing Spec files.

klihub:
Add convenience functions for generating Spec names for ordinary and transient Spec files. The generated names can be passed to WriteSpec() of Cache or Registry to atomically write a Spec file. Also add a function for removing Spec files created with WriteSpec().

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -255,11 +255,10 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	return nil, nil
 }
 
-// WriteSpec writes a Spec file with the given content. Priority is used
-// as an index into the list of Spec directories to pick a directory for
-// the file, adjusting for any under- or overflows. If name has a "json"
-// or "yaml" extension it choses the encoding. Otherwise JSON encoding
-// is used with a "json" extension.
+// WriteSpec writes a Spec file with the given content into the highest
+// priority Spec directory. If name has a "json" or "yaml" extension it
+// choses the encoding. Otherwise JSON encoding is used with a "json"
+// extension appended.
 func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 	var (
 		specDir string
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -17,10 +17,14 @@
 package cdi
 
 import (
+	"io/fs"
+	"os"
 	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
+
+	stderr "errors"
 
 	cdi "github.com/container-orchestrated-devices/container-device-interface/specs-go"
 	"github.com/fsnotify/fsnotify"
@@ -255,6 +259,19 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	return nil, nil
 }
 
+// highestPrioritySpecDir returns the Spec directory with highest priority
+// and its priority.
+func (c *Cache) highestPrioritySpecDir() (string, int) {
+	if len(c.specDirs) == 0 {
+		return "", -1
+	}
+
+	prio := len(c.specDirs) - 1
+	dir := c.specDirs[prio]
+
+	return dir, prio
+}
+
 // WriteSpec writes a Spec file with the given content into the highest
 // priority Spec directory. If name has a "json" or "yaml" extension it
 // choses the encoding. Otherwise JSON encoding is used with a "json"
@@ -268,12 +285,11 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 		err     error
 	)
 
-	if len(c.specDirs) == 0 {
+	specDir, prio = c.highestPrioritySpecDir()
+	if specDir == "" {
 		return errors.New("no Spec directories to write to")
 	}
 
-	prio = len(c.specDirs) - 1
-	specDir = c.specDirs[prio]
 	path = filepath.Join(specDir, name)
 	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
 		path += ".json"
@@ -285,6 +301,35 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 	}
 
 	return spec.Write(true)
+}
+
+// RemoveSpec removes a Spec with the given name from the highest
+// priority Spec directory. This function can be used to remove a
+// Spec previously written by WriteSpec(). If the file exists and
+// its removal fails RemoveSpec returns an error.
+func (c *Cache) RemoveSpec(name string) error {
+	var (
+		specDir string
+		path    string
+		err     error
+	)
+
+	specDir, _ = c.highestPrioritySpecDir()
+	if specDir == "" {
+		return errors.New("no Spec directories to remove from")
+	}
+
+	path = filepath.Join(specDir, name)
+	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
+		path += ".json"
+	}
+
+	err = os.Remove(path)
+	if err != nil && stderr.Is(err, fs.ErrNotExist) {
+		err = nil
+	}
+
+	return err
 }
 
 // GetDevice returns the cached device for the given qualified name.
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -274,8 +274,7 @@ func (c *Cache) highestPrioritySpecDir() (string, int) {
 
 // WriteSpec writes a Spec file with the given content into the highest
 // priority Spec directory. If name has a "json" or "yaml" extension it
-// choses the encoding. Otherwise JSON encoding is used with a "json"
-// extension appended.
+// choses the encoding. Otherwise the default YAML encoding is used.
 func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 	var (
 		specDir string
@@ -292,7 +291,7 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 
 	path = filepath.Join(specDir, name)
 	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
-		path += ".json"
+		path += defaultSpecExt
 	}
 
 	spec, err = NewSpec(raw, path, prio)
@@ -321,7 +320,7 @@ func (c *Cache) RemoveSpec(name string) error {
 
 	path = filepath.Join(specDir, name)
 	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
-		path += ".json"
+		path += defaultSpecExt
 	}
 
 	err = os.Remove(path)
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -294,12 +294,12 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
 		path += defaultSpecExt
 	}
 
-	spec, err = NewSpec(raw, path, prio)
+	spec, err = newSpec(raw, path, prio)
 	if err != nil {
 		return err
 	}
 
-	return spec.Write(true)
+	return spec.write(true)
 }
 
 // RemoveSpec removes a Spec with the given name from the highest
--- a/pkg/cdi/cache_test.go
+++ b/pkg/cdi/cache_test.go
@@ -17,8 +17,10 @@
 package cdi
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
@@ -29,6 +31,7 @@ import (
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/yaml"
 )
 
 func TestNewCache(t *testing.T) {
@@ -1628,6 +1631,215 @@ containerEdits:
 					name := filepath.Base(spec.GetPath())
 					require.Equal(t, spec.Spec, cSpecs[name])
 				}
+			}
+		})
+	}
+}
+
+func TestCacheTransientSpecs(t *testing.T) {
+	type testCase struct {
+		name         string
+		specs        []string
+		invalid      map[int]bool
+		expected     [][]string
+		numSpecFiles []int
+	}
+	for _, tc := range []*testCase{
+		{
+			name: "invalid spec",
+			specs: []string{
+				`
+cdiVersion: "` + cdi.CurrentVersion + `"
+kind:       "vendor.comdevice"
+devices:
+  - name: "dev1"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor1-dev1"
+        type: b
+        major: 10
+        minor: 1`,
+			},
+			invalid: map[int]bool{
+				0: true,
+			},
+		},
+		{
+			name: "add/remove one valid spec",
+			specs: []string{
+				`
+cdiVersion: "` + cdi.CurrentVersion + `"
+kind:       "vendor.com/device"
+devices:
+  - name: "dev1"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor-dev1"
+        type: b
+        major: 10
+        minor: 1
+`,
+				"-0",
+			},
+			expected: [][]string{
+				[]string{
+					"vendor.com/device=dev1",
+				},
+				nil,
+			},
+			numSpecFiles: []int{
+				1,
+				0,
+			},
+		},
+		{
+			name: "add/remove multiple valid specs",
+			specs: []string{
+				`
+cdiVersion: "` + cdi.CurrentVersion + `"
+kind:       "vendor.com/device"
+devices:
+  - name: "dev1"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor-dev1"
+        type: b
+        major: 10
+        minor: 1
+`,
+				`
+cdiVersion: "` + cdi.CurrentVersion + `"
+kind:       "vendor.com/device"
+devices:
+  - name: "dev2"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor-dev2"
+        type: b
+        major: 10
+        minor: 2
+`,
+				`
+cdiVersion: "` + cdi.CurrentVersion + `"
+kind:       "vendor.com/device"
+devices:
+  - name: "dev3"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor-dev3"
+        type: b
+        major: 10
+        minor: 3
+  - name: "dev4"
+    containerEdits:
+      deviceNodes:
+      - path: "/dev/vendor-dev4"
+        type: b
+        major: 10
+        minor: 4
+`,
+				"-0",
+				"-1",
+				"-2",
+			},
+			expected: [][]string{
+				[]string{
+					"vendor.com/device=dev1",
+				},
+				[]string{
+					"vendor.com/device=dev1",
+					"vendor.com/device=dev2",
+				},
+				[]string{
+					"vendor.com/device=dev1",
+					"vendor.com/device=dev2",
+					"vendor.com/device=dev3",
+					"vendor.com/device=dev4",
+				},
+				[]string{
+					"vendor.com/device=dev2",
+					"vendor.com/device=dev3",
+					"vendor.com/device=dev4",
+				},
+				[]string{
+					"vendor.com/device=dev3",
+					"vendor.com/device=dev4",
+				},
+				nil,
+			},
+			numSpecFiles: []int{
+				1,
+				2,
+				3,
+				2,
+				1,
+				0,
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var (
+				dir       string
+				err       error
+				cache     *Cache
+				specFiles []os.DirEntry
+				specs     = map[int]string{}
+			)
+
+			dir, err = createSpecDirs(t, nil, nil)
+			require.NoError(t, err)
+			cache, err = NewCache(
+				WithSpecDirs(
+					filepath.Join(dir, "etc"),
+					filepath.Join(dir, "run"),
+				),
+				WithAutoRefresh(false),
+			)
+
+			require.NoError(t, err)
+			require.NotNil(t, cache)
+
+			for idx, data := range tc.specs {
+				var (
+					transientID string
+					raw         *cdi.Spec
+					delIdx      int
+					err         error
+				)
+
+				if data[0] == '-' {
+					delIdx, err = strconv.Atoi(string(data[1:]))
+					require.NoError(t, err)
+
+					err = cache.RemoveSpec(specs[delIdx])
+					require.NoError(t, err)
+				} else {
+					err = yaml.Unmarshal([]byte(data), &raw)
+					require.NoError(t, err)
+
+					transientID = fmt.Sprintf("id%d", idx)
+					specs[idx], err = GenerateNameForTransientSpec(raw, transientID)
+					if tc.invalid[idx] {
+						require.NotNil(t, err)
+						continue
+					}
+					require.NoError(t, err)
+
+					err = cache.WriteSpec(raw, specs[idx])
+					require.NoError(t, err)
+				}
+
+				err = cache.Refresh()
+				require.NoError(t, err)
+
+				devices := cache.ListDevices()
+				require.Equal(t, tc.expected[idx], devices)
+
+				specFiles, err = os.ReadDir(
+					filepath.Join(dir, "run"),
+				)
+				require.NoError(t, err)
+				require.Equal(t, tc.numSpecFiles[idx], len(specFiles))
 			}
 		})
 	}
--- a/pkg/cdi/doc.go
+++ b/pkg/cdi/doc.go
@@ -180,7 +180,7 @@
 // func generateDeviceSpecs() error {
 //     registry := cdi.GetRegistry()
 //     spec := &specs.Spec{
-//         Version: specVersion,
+//         Version: specs.CurrentVersion,
 //         Kind:    vendor+"/"+class,
 //     }
 //
@@ -216,7 +216,7 @@
 //     registry := cdi.GetRegistry()
 //     devices := getContainerDevs(ctr, vendor, class)
 //     spec := &specs.Spec{
-//         Version: specVersion,
+//         Version: specs.CurrentVersion,
 //         Kind:    vendor+"/"+class,
 //     }
 //
--- a/pkg/cdi/doc.go
+++ b/pkg/cdi/doc.go
@@ -124,10 +124,15 @@
 //
 // Generated Spec Files, Multiple Directories, Device Precedence
 //
-// There are systems where the set of available or usable CDI devices
-// changes dynamically and this needs to be reflected in the CDI Specs.
-// This is done by dynamically regenerating CDI Spec files which are
-// affected by these changes.
+// It is often necessary to generate Spec files dynamically. On some
+// systems the available or usable set of CDI devices might change
+// dynamically which then needs to be reflected in CDI Specs. For
+// some device classes it makes sense to enumerate the available
+// devices at every boot and generate Spec file entries for each
+// device found. Some CDI devices might need special client- or
+// request-specific configuration which can only be fulfilled by
+// dynamically generated client-specific entries in transient Spec
+// files.
 //
 // CDI can collect Spec files from multiple directories. Spec files are
 // automatically assigned priorities according to which directory they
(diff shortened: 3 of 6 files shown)


klihub:
@bart0sh @klueska I cooked up this patch set for dealing with consistently named vendor-specific dynamically generated Spec files. Using the added two functions should guarantee global uniqueness of the generated Spec files with atomic updates to the cache/registry from the Refresh() point of view.

If this is found to be a usable abstraction then I'd also like to replace the current WriteSpec() with a corresponding pair of CreateSpec()/RemoveSpec(). These would provide atomic updates and use a similar vendor+device-namespaced naming convention but with reproducible file names (and take an overwrite flag like WriteSpec()).

pkg/cdi/cache.go
```diff
@@ -288,6 +312,126 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
...
+// If a non-empty optional tag is given, it is made a distinctive part of
+// the generated file name. This can be used to help identify Spec sources
+// in cases where multiple clients generate transient Specs with identical
+// vendor and class, simply by using a unique tag per client.
+//
+// A transient Spec is one whose lifecycle is tied to that of an external
+// entity or to events external to CDI and CDI clients, for instance the
+// creation and removal of a particular container. Generated transient Spec
+// files should eventually be removed using RemoveTransientSpec().
+func (c *Cache) CreateTransientSpec(raw *cdi.Spec, tag string) (path string, err error) {
```
pohly, line 331: There are two problems with using this API instead of a deterministic name:
- the generated name must be stored somewhere
- creating the file must be idempotent to handle the case that the process gets killed right after creating the file and before doing anything further.

With a deterministic file name that's easy (just overwrite any existing file with that name, delete the file with the name if it still exists).

With this helper function the caller would first need to read all spec to check whether the content is already there, then only write another transient file if not.

I prefer defining a naming convention for transient files that all vendors are expected to follow.

klihub: > There are two problems with using this API instead of a deterministic name:
> 
> * the generated name must be stored somewhere
> * creating the file must be idempotent to handle the case that the process gets killed right after creating the file and before doing anything further.
> 
> With a deterministic file name that's easy (just overwrite any existing file with that name, delete the file with the name if it still exists).
> 
> With this helper function the caller would first need to read all spec to check whether the content is already there, then only write another transient file if not.
> 
> I prefer defining a naming convention for transient files that all vendors are expected to follow.

Yes with this API a crash at that point becomes a real immediate problem, unless the restarted client generates the Spec file again with different names referencing the container-specific CDI devices... and we probably can't trust the driver to do that.

Then we could go with something like this:

```
    // GetSpecName generates vendor+class-scoped spec file names.
    // If a non-empty driver is given, it should uniquely identify the caller among multiple possible
    // CDI users on the same host generating Spec files for the same vendor/class. This function
    // is idempotent. It returns the same name if called with the same arguments.
    func GetSpecName(vendor, class, driver string) string {
        // ...
    }

    // GetTransientSpecName generates vendor+class-scoped transient spec file names.
    // If a non-empty driver is given, it should uniquely identify the caller among multiple possible
    // CDI users on the same host generating Spec files for the same vendor/class. Similarly, if a
    // non-empty transientID is given, it should uniquely identify an external entity to which the
    // Spec file corresponds to. This can be for instance the ID of a container for which the Spec
    // entries are exclusively generated. This function is not idempotent. It returns a unique name
    // on each invocation.
    func GetTransientSpecName(vendor, class, driver string) string {
        // ...
    }

    // Spec files are written with the already existing WriteSpec()...
```

This separates file name generation from Spec file writing and provides separate convenience functions for transient and other Spec file name generation. Transient file names are internally are generated unique per call. The client has now a chance to store the file name before writing the actual file (using WriteSpec()). If the DRA device driver/plugin crashes at the critical/problematic point, it can just reused the already generated name if it has one, or generate a name if it did not manage to store one yet.

I assume that the DRA device driver/plugin is already stateful so storing state is not a problem. If the driver is stateless or storing a file name is a prohibitively insurmountable problem, an alternative would be combining these to a single function

```
    // GetSpecName generates vendor+class-scoped spec file names.
    // If a non-empty driver is given, it should uniquely identify the caller among multiple possible
    // CDI users on the same host generating Spec files for the same vendor/class. Similarly, if a
    // non-empty transientID is given, it should uniquely identify an external entity to which the
    // Spec file corresponds to. This can be for instance the ID of a container for which the Spec
    // entries are exclusively generated.
    func GetSpecName(vendor, class, driver, transientID string) string {
        // ...
    }
```

In this case we can't really guarantee uniqueness which was the main point of these additions. We have to rely in the caller to do the right thing based on the documented usage of the API function.

pkg/cdi/cache.go
```diff
@@ -17,18 +17,31 @@
...
 	"github.com/pkg/errors"
 )
 
+const (
+	// reservationDir is the subdirectory for transient Spec reservations.
+	reservationDir = ".transient"
+	// transientPrefix is the base prefix used to transient Spec files.
+	transientPrefix = "transient."
+	// transientExt is the type/extension for transient Spec files.
+	transientExt = ".json"
```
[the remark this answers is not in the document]
bart0sh: +1 for using YAML. JSON is good as a transport protocol in my opinion. YAML is more suitable for storing configuration info and extensively used in k8s for this purpose.

pkg/cdi/cache.go
bart0sh: Currently example DRA driver is stateless and it could be the case for real drivers as well. We ended up using CDI file path /var/run/cdi/<driver name>-<resource claim UID>.json as example driver name(default: example-driver.cdi.k8s.io) contains vendor domain in our case, so using both vendor and driver would look tautological.

bart0sh: From other point of view it's not a big deal for the driver to refresh cdi cache on its start, which should give the driver all information it needs.

bart0sh: @pohly > With this helper function the caller would first need to read all spec to check whether the content is already there, then only write another transient file if not.

This is done by cdi API as well as far as I know.

bart0sh: @klueska your input here would be appreciated as you've written first(and the only existing for now) real DRA plugin.

pohly: A helper function like GetSpecName would be useful. "vendor" and "class" seem a bit redundant when "driver" alone is already required to be unique, which is the case, right?

pkg/cdi/cache.go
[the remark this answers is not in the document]
klihub: I'd also prefer YAML. I went with JSON because I thought there is a strong preference for it (the original CDI implementation did not support YAML at all). Is there a need to let the user choose the format for these generated files or can we always go with YAML here ?

pkg/cdi/cache.go
klihub: No, in the scheme I sketched up the driver was an optional(ly non-empty) parameter. It is supposed to be unique only within the same vendor/class domain. It's there to allow multiple drivers that handle devices with identical vendor and class to coexist on the same node, yet use this function to generate non-conflicting file names.

For instance, if a vendor had one driver dealing with legacy HW and another dealing with recent ones, then the legacy could use "legacy" as the driver and the other could use "" (or "current" or something equivalent).

klihub: > @pohly > With this helper function the caller would first need to read all spec to check whether the content is already there, then only write another transient file if not.
> 
> This is done by the CDI API as far as I know.

@bart0sh Do you mean that your example driver is doing this on the client side ?

bart0sh: Current implementation of the example driver just creates a spec, marshals it into json and writes to /var/run/cdi/<driver name>-<resource claim UID>.json
However, it can use CDI cache API on the start to get all CDI files parsed, stored in the cache and monitored I guess.

pkg/cdi/cache.go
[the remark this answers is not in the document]
bart0sh: It would be better to let the user choose the format, but default should be YAML I think.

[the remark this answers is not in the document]
bart0sh: YAML is also preferable considering @klueska's opinion that most of plugins would end up using static CDI files. His NVIDIA GPU DRA plugin only uses static files.

pkg/cdi/cache.go
klihub: > Current implementation of the example driver just creates a spec, marshals it into json and writes to `/var/run/cdi/<driver name>-<resource claim UID>.json` However, it can use CDI cache API on the start to get all CDI files parsed, stored in the cache and monitored I guess.

Ah, okay... you meant that. I think we'd like to avoid such extra complexity in the driver, ideally keeping it as simple as `generate Spec content + GetSpecName() + WriteSpec()`. So IOW keeping it like you have it now, just using CDI-provided convenience functions to generate the Spec name (according to our preferred convention) and to write the content (making sure the changes are atomic wrt. Refresh()). Then the example driver would serve as a properly written template for driver writers to copy and modify.

klueska:
I only took a high-level look at the code, but in general, I agree with the other sentiments given.

Rather than create / delete files through the CDI library, we should add functions to:
1. Generate a transient / unique file name following a specific convention
2. Generate a transient / unique device name following a specific convention

It is then up to the user to call these functions to generate these transient filenames / device names, but other than that, nothing changes about the way the current API is used to create a spec / write a spec file out to disk.

klihub:
So how about something like this. This is the same as the combined single function above, but it does not hardcode JSON as the file format. I tried to formulate the documentation of how to use it, including the driver and transientID .

```
// GetSpecName generates a vendor+class scoped Spec file name. The name
// can be passed to WriteSpec() to write a Spec file to the file system.
//
// The file name is generated without a ".json" or ".yaml" extension.
// The caller can append the desired extension to choose a particular
// encoding. Otherwise WriteSpec() will use its default encoding.
//
// vendor and class should match the vendor and class of the CDI Spec.
// driver can be left empty if only a single entity is generating Spec
// files for the given vendor/class combination on the host. Otherwise
// driver should uniquely identify the caller among the multiple ones
// that generate Spec files with the same vendor and class.
//
// If the lifecycle of the Spec file is tied to the lifecycle of some
// entity external to CDI then a non-empty transientID should be used.
// This should uniquely identify the external entity. For instance, it
// can be the ID of a container, if the Spec will be generated for use
// by a single container.
func GetSpecName(vendor, class, driver, transientID string) string {
  // ...
}
```

We could also use the same Spec the client will pass to WriteSpec() to dig out vendor/class ourselves. With that the signature would be something like `func GetSpecName(spec *cdi.Spec, driver, transientID string) string`.

pkg/cdi/cache.go
klihub: > One thing to keep in mind regarding the following:
> 
> > Then the example driver would serve as a properly written template for driver writers to copy and modify.
> 
> Writing transient spec files like this is likely _not_ going to be done by many (if any) actual driver implementors. I see it as the exception rather than the common case. Most people will write one spec file once (at machine bootup or at driver start) and then refer to the various devices defined within it as allocation requests come in.
> 
> The whole idea of generating specs on the fly was new to me when I looked at the example driver, and it is definitely _not_ something we plan to do in the NVIDIA driver.

If that is true, then wouldn't it make better sense to provide *separate* functions for transient and 'ordinary' Spec file name generation ? For instance `GetSpecName()` and  `GetTransientSpecName()`. Then `driver` and `transientID` would only be passed to the transient one, so most people would not need to even wonder about what they are and how to correctly pick them.

pohly: > I think we'd like to avoid such extra complexity in the driver, ideally keeping it as simple as generate Spec content + GetSpecName() + WriteSpec()

Agreed.

> Writing transient spec files like this is likely not going to be done by many (if any) actual driver implementors.

That depends. If the driver needs some actual input from the user before preparing the hardware (like programming an FPGA) and that then affects how the hardware is presented in the container, then the CDI file has to be generated.

> GenerateSpecName() and GenerateTransientSpecName()

:+1:

bart0sh: @klihub > generate Spec content + GetSpecName() + WriteSpec()

 I'm wondering if user even need to know the CDI file name? It should be possible to get it when CDI spec is created, but getting the name before writing the spec to the file might not be needed at all. Would it be enough in most cases to generate Spec content (spec := NewSpec(...))  + spec.Write() ?

pohly: > I'm wondering if user even need to know the CDI file name?

Yes, for (debug) log messages.

pohly: We also need the name for removing the file.

bart0sh: > Yes, for (debug) log messages

spec.GetPath() should be enough for that.

bart0sh: What I was trying to say is that user of the spec API might not be interested in knowing the file name at all. What's important for user is to know that spec is stored somewhere and the runtime can access it.

klihub: > @klihub > generate Spec content + GetSpecName() + WriteSpec()
> 
> I'm wondering if user even need to know the CDI file name? It should be possible to get it when CDI spec is created (spec.GetPath()), but getting the name before writing the spec to the file might not be needed at all. Would it be enough in most cases to generate Spec content (spec := NewSpec(...)) + writing the spec (spec.Write()) ?

Hmm... if we this route that @pohly and @klueska seem to prefer then we'll need to add either `RemoveSpec(vendor, class), RemoveTransientSpec(vendor, class, driver, transientID)` or just a `RemoveSpec(name)`, because the generated name is not a path just a basename. It is `WriteSpec()` which picks the last spec directory and writes the Spec there, using the given name, also chosing a default encoding if necessary.

There would be other ways to do that (look up the Spec using the unique device name, get the path of that spec, the os.Remove() it). But if there is implicit logic/assumptions for Spec creation in the package, then it should rather provide functions with the same implicit knowledge/assumption for Spec removal.

bart0sh: Correct, but not only that. CDI should be able to also understand if a spec really needs to be written, i.e. it should compare desired spec content with the content already present in the file/cache. If content is not changed, we should avoid writing it again. This can help avoiding races between driver writing the spec and runtime reading it at the same time.

This may be an overcomplication on my side and simple atomic writing should be enough. As far as I can see in [the code](https://github.com/container-orchestrated-devices/container-device-interface/blob/main/pkg/cdi/spec.go#L154) spec.Write() writes CDI file atomically.

pkg/cdi/spec.go
```diff
@@ -262,3 +262,56 @@ func validateSpec(raw *cdi.Spec) error {
...
+// The caller can append the desired extension to choose a particular
+// encoding. Otherwise WriteSpec() will use its default encoding.
+func GenerateTransientSpecName(vendor, class, driver, transientID string) string {
+	name := vendor + "-" + class
+
+	if driver != "" {
+		name += "+" + driver
+	}
+
+	name = "transient." + name + "_" + transientID
```
bart0sh, line 314: Would it be better to always start file names with vendor and class so that they're nicely grouped in the ls output?

klihub: True. I'll update it like that.

pkg/cdi/cache.go
```diff
@@ -288,6 +303,35 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
...
+// Spec previosuly written by WriteSpec() using the same. If the
+// file exists and its removal fails RemoveSpec returns an error.
+func (c *Cache) RemoveSpec(name string) error {
+	var (
+		specDir string
+		path    string
+		err     error
+	)
+
+	specDir, _ = c.highestPrioritySpecDir()
```
bart0sh, line 317: Does it mean that RemoveSpec could only remove CDI files from the directory with the highest priority?

pkg/cdi/spec.go
```diff
@@ -262,3 +262,75 @@ func validateSpec(raw *cdi.Spec) error {
...
+// if only a single CDI user is generating transient Spec files for the
+// given vendor/class combination on the host. Otherwise driver should
+// uniquely identify the caller among those that use the same vendor
+// and class. transientID should uniquely identify the external entity
+// within the scope of the caller/driver.
+//
+// The file name is generated without a ".json" or ".yaml" extension.
+// The caller can append the desired extension to choose a particular
+// encoding. Otherwise WriteSpec() will use its default encoding.
+func GenerateTransientSpecName(vendor, class, driver, transientID string) string {
```
bart0sh, line 298: What if we remove driver from here? Caller can include it into the transientID if needed. Otherwise we'd need to explain what is it, which just seems not needed for me. Any software can call this API, not just drivers or plugins.

pkg/cdi/cache.go
klihub: Yes. It is the only/same directory where WriteSpec() writes Spec files, so RemoveSpec() removes only from that same directory.

pkg/cdi/spec.go
klihub: I'm fine with that if we are sure that this is not going to cause any problems (in the case when multiple drivers/plugins generate Spec files with the same vendor/class combination).

pkg/cdi/cache.go
bart0sh: So, not every spec in cache can be removed using RemoveSpec, but only those that are located in the directory with the highest priority?

pkg/cdi/cache.go
klueska: > > Writing transient spec files like this is likely not going to be done by many (if any) actual driver implementors.
> 
> That depends. If the driver needs some actual input from the user before preparing the hardware (like programming an FPGA) and that then affects how the hardware is presented in the container, then the CDI file has to be generated.

Our plan for this is to have separate „devices“ predeclared for these dynamic options that declare what needs to be injected base on configuring then device this way. This won’t work for all use cases but it will work for us.

pkg/cdi/cache.go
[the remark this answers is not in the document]
klihub: Changed the default encoding to YAML.

pkg/cdi/cache.go
klihub: > Correct, but not only that. CDI should be able to also understand if a spec really needs to be written, i.e. it should compare desired spec content with the content already present in the file/cache. If content is not changed, we should avoid writing it again.

I'm not sure if any of the golang JSON encoders are stable (producing identical output for semantically identical input). I recall the stock one is not. So let's leave this as a potential future exercise/improvement.

bart0sh: works for me

bart0sh:
@klihub As far as I understood there are 3 ways of writing spec to the file:
spec.Write(overwrite bool)
registry.WriteSpec(name string)
cache.WriteSpec(cache.raw *cdi.Spec, name string)

Do we really need all 3 of them?
Can you describe possible usage scenarios and recommendations for each?

klihub:
> @klihub As far as I understood there are 2 ways of writing spec to the file: spec.Write(overwrite bool) cache.WriteSpec(cache.raw *cdi.Spec, name string)
> 
> Can you describe possible usage scenarios and recommendations for each?

Unless you are doing something special which requires you to fully control the exact path where the Spec file gets written to, you should always just use WriteSpec(). And as I mention already earlier, it might be better to change Spec.Write() into spec.write() (IOW, make it package private function).

klihub:
> > @klihub As far as I understood there are 2 ways of writing spec to the file: spec.Write(overwrite bool) cache.WriteSpec(cache.raw *cdi.Spec, name string)
> > Can you describe possible usage scenarios and recommendations for each?
> 
> Unless you are doing something special which requires you to fully control the exact path where the Spec file gets written to, you should always just use WriteSpec(). And as I mentioned already earlier, it might be better to change Spec.Write() into spec.write() (IOW, make it package private function) to eliminate any confusion which one should be used.

pkg/cdi/doc.go
```diff
@@ -141,7 +146,111 @@
...
+// )
+//
+// func generateDeviceSpecs() error {
+//     registry := cdi.GetRegistry()
+//     spec := &specs.Spec{
+//         Version: specVersion,
+//         Kind:    vendor+"/"+class,
+//     }
+//
+//     for _, dev := range enumarateDevices() {
```
bart0sh, line 187: typo: enum**a**rateDevices()

pkg/cdi/doc.go
```diff
@@ -141,7 +146,111 @@
...
+//         return fmt.Errorf("failed to generate Spec name: %w", err)
+//     }
+//
+//     return registry.WriteSpec(spec, specName)
+// }
+//
+// Similary, generating and later cleaning up transient Spec files can be
+// done with code fragments similar to the following. These transient Spec
+// files are temporary Spec files with container-specific parametrization.
+// They are typically created created before the associated container is
```
bart0sh, line 205: typo: created created

pkg/cdi/spec.go
```diff
@@ -262,3 +271,67 @@ func validateSpec(raw *cdi.Spec) error {
...
+// transient Spec files using the same vendor/class combination. If
+// the external entity to which the lifecycle of the tranient Spec
+// is tied to has a unique ID of its own, then this is usually a
+// good choice for transientID.
+//
+// The file name is generated without a ".json" or ".yaml" extension.
+// The caller can append the desired extension to choose a particular
+// encoding. Otherwise WriteSpec() will use its default encoding.
+func GenerateTransientSpecName(vendor, class, transientID string) string {
+	return vendor + "-" + class + ".transient_" + transientID
```
bart0sh, line 308: I'd suggest to remove ".transient_" part. This would give users of this API more control over the name.

bart0sh:
@klihub the PR looks good to me. I made some nit-picking comments, but the overall impression is good. Thank you!

klihub:
> @klihub the PR looks good to me. I made some nit-picking comments, but the overall impression is good. Thank you!

Thanks ! Those are all valid. I'll address/fix them once I'm back from lunch.

klihub:
> > @klihub the PR looks good to me. I made some nit-picking comments, but the overall impression is good. Thank you!
> 
> Thanks ! Those are all valid. I'll address/fix them once I'm back from lunch.

@bart0sh Done.

pkg/cdi/spec.go
```diff
@@ -262,3 +272,69 @@ func validateSpec(raw *cdi.Spec) error {
...
+// GenerateNameForTransientSpec generates a name for the given transient
+// Spec using GenerateTransientSpecName with the vendor and class taken
+// from the Spec. On success it returns the generated name and a nil error.
+// If the Spec does not contain a valid vendor or class, it returns an
+// an empty name and a non-nil error.
+func GenerateNameForTransientSpec(raw *cdi.Spec, transientID string) (string, error) {
+	vendor, class := ParseQualifier(raw.Kind)
+	if vendor == "" {
+		return "", errors.Errorf("invalid vendor/class %q in Spec", raw.Kind)
+	}
```
pohly, line 336: Note that the test driver cannot use any of these functions because the kubelet plugin might run inside the e2e.test binary (i.e. outside of the cluster) and just manipulate the nodes from remote (using port forwarding and the equivalent of `kubectl exec dd/rm` to create and remove files) - see https://github.com/pohly/kubernetes/pull/15 and in particular the IO abstraction in [test/e2e/dra/test-driver/app/kubeletplugin.go](https://github.com/pohly/kubernetes/pull/15/commits/dba16e41b7590d886c76a9b113449677c8d3eb85#diff-f20c62b4b40543cf14c979f85bfe98c0ab0e2b1b714458423d64387977df2507).

The API for generating the names might be useful for the test driver if it included the full path, but not as it stands right now because only WriteSpec internally creates the full path.

bart0sh (approved):
/lgtm

pkg/cdi/cache_test.go
```diff
@@ -1634,6 +1637,215 @@ containerEdits:
...
+		invalid      map[int]bool
+		expected     [][]string
+		numSpecFiles []int
+	}
+	for _, tc := range []*testCase{
+		{
+			name: "invalid spec",
+			specs: []string{
+				`
+cdiVersion: "0.3.0"
```
elezar, line 1653: Question: SHould we use the latests spec version for tests like this?

pkg/cdi/spec.go
bart0sh: @klihub would it make sense to add an API to get the file path for the written spec? This would allow test driver to get the file path and transfer the file to the remote location.

bart0sh: Alternatively Cache.WriteSpec can return path to the written spec. @pohly would it help test driver use case?

pkg/cdi/cache_test.go
klihub: Good point. I updated it like that.

pkg/cdi/spec.go
klihub: @bart0sh We can update `WriteSpec()` to return `(path string, err error)` if that helps.

bart0sh: @pohly  wdyt?

pohly: The test driver cannot call `WriteSpec`. That would write the file on the host where the e2e.test runs, not on the host where kubelet and the container runtime run.

bart0sh: Can the e2e transfer the file to the host where runtime can access it?

pohly: Yes. But it doesn't have write access to where this package wants to write the file (`/var/run/cdi`), so writing it first locally and then copying to the remote host won't work.

bart0sh: @pohly the code that creates a CDI file is [here](https://github.com/pohly/kubernetes/blob/dynamic-resource-allocation/test/e2e/dra/test-driver/app/kubeletplugin.go#L125:L157) as far as I understand. It still creates a file on the file system. Can you point us out to the  e2e code that uses this file(transfers it to a remote location?)?

pohly: > It still creates a file on the file system.

No. It encodes the spec in memory (`buffer, err := json.Marshal(*cdiSpec)`) then creates the file on the remote remote host (`ex.fileOps.Create(filePath, buffer)`). The implementation of that file creation is in https://github.com/pohly/kubernetes/blob/c15323637d09a28d7100527a38a2daf3e20e9746/test/e2e/storage/drivers/proxy/io.go#L65-L71

bart0sh: @pohly Thanks. Now I more or less understand how it works.

With the proposed API change it will not be possible I guess as newSpec is a private method there.
So, we need to either leave NewSpec public or modify an API that it can be used by e2e.

pohly (commented):
Just one small nit.

However, my earlier comment regarding these new functions still stands: because the library assumes that it has direct access to the directories, it cannot be used for the Kubernetes test/e2e/dra/test-driver where all of the logic runs outside of the cluster and only indirectly manipulates files.

pkg/cdi/doc.go
```diff
@@ -141,7 +146,111 @@
...
+//     "fmt"
+//     ...
+//     "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+//     "github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
+// )
+//
+// func generateDeviceSpecs() error {
+//     registry := cdi.GetRegistry()
+//     spec := &specs.Spec{
+//         Version: specVersion,
```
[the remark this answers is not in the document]
pohly: You later commit adds an alias for `cdi.SpecVersion`. Would be good to use that alias here.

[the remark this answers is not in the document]
bart0sh: @klihub can you address this? It seems to be the last comment. I hope that as soon as this comment is resolved we can merge this PR and make a new CDI release.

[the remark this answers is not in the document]
klihub: @bart0sh I changed the tip-most commit to update the code same in `doc.go` to use the newly exposed `specs.CurrentVersion` as the generated version of the spec (which I what I'd expect/like folks to do by default: generate/use specs with the latest available version of the package they compile against).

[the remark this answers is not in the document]
klihub: > You later commit adds an alias for `cdi.SpecVersion`. Would be good to use that alias here.

Good point and sorry for having missed this originally and now for so long. Updated the topmost commit of PR exactly like that.

bart0sh:
@klueska @pohly @elezar all comments have been addressed. Can you review again and merge?

bart0sh:
@elezar would it make sense to make a new release after merging this? I hope I can update CDI in containerd before their release if we release fast enough.

pohly:
> I hope I can update CDI in containerd before their release if we release fast enough.

Does this change matter for containerd?

bart0sh:
This change doesn't matter for containerd. However, I'd prefer to update CDI there to the latest possible CDI release.

pkg/cdi/spec.go
```diff
@@ -270,3 +283,69 @@ func validateSpec(raw *cdi.Spec) error {
...
+
+	transientID = strings.ReplaceAll(transientID, "/", "_")
```
elezar, line 349: A minor nit, but isn't this already done in `GenerateTransientSpecName`?

pkg/cdi/spec.go
```diff
@@ -270,3 +283,69 @@ func validateSpec(raw *cdi.Spec) error {
...
+// the external entity to which the lifecycle of the tranient Spec
+// is tied to has a unique ID of its own, then this is usually a
+// good choice for transientID.
+//
+// The file name is generated without a ".json" or ".yaml" extension.
+// The caller can append the desired extension to choose a particular
+// encoding. Otherwise WriteSpec() will use its default encoding.
+func GenerateTransientSpecName(vendor, class, transientID string) string {
+	transientID = strings.ReplaceAll(transientID, "/", "_")
+	return vendor + "-" + class + "_" + transientID
```
elezar, line 321: Nit: Would calling `GenerateSpecName(vendor, class)` here be more consistent?

elezar (approved):
@klihub I think this looks good from my side. I'll just give @klueska @pohly and @bart0sh a moment to respond before merging.

elezar:
> @elezar would it make sense to make a new release after merging this? I hope I can update CDI in containerd before their release if we release fast enough.

@bart0sh sure. We can cut a new release as soon as possible after this has landed. The operation is lightweight enough and we don't expect any breaking changes from automatic updates by consumers.

pkg/cdi/doc.go
```diff
@@ -141,7 +146,111 @@
...
+//         return fmt.Errorf("failed to generate Spec name: %w", err)
+//     }
+//
+//     return registry.WriteSpec(spec, specName)
+// }
+//
+// func removeTransientSpec(ctr Container) error {
+//     registry := cdi.GetRegistry()
+//     transientID := getSomeUniqueIDForContainer(ctr)
+//     specName := cdi.GenerateTransientSpecName(vendor, class, transientID)
```
klueska, line 250: Lines 238 and 239 should match lines 249 and 250, right?

pkg/cdi/spec.go
klihub: True, removed.

pkg/cdi/spec.go
klihub: Updated to use that.

pkg/cdi/cache.go
```diff
@@ -269,23 +284,51 @@ func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
...
-	if len(c.specDirs) == 0 {
+	specDir, prio = c.highestPrioritySpecDir()
+	if specDir == "" {
 		return errors.New("no Spec directories to write to")
 	}
 
-	prio = len(c.specDirs) - 1
-	specDir = c.specDirs[prio]
 	path = filepath.Join(specDir, name)
 	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
```
elezar, line 293: This is already set in `newSpec`, correct? So we could drop this entire conditional.

elezar (approved):
Thanks again for the work here @klihub.

I think we can merge this and then address any minor points in a follow-up.

pohly:
I updated to the new CDI release and this PR broke how the DRA test driver generates specs:

```
spec := &specs.Spec{ ... }
cdiSpec, err := cdiapi.NewSpec(&spec, filePath, 1)
```

What is the right way of creating a cdiapi.Spec instance now?

I need it further down for:
```
dev := cdiSpec.GetDevice(deviceName).GetQualifiedName()
resp := &drapbv1.NodePrepareResourceResponse{CdiDevices: []string{dev}}
```

pohly:
I tried this instead, but tests didn't pass:
```
diff --git a/test/e2e/dra/test-driver/app/kubeletplugin.go b/test/e2e/dra/test-driver/app/kubeletplugin.go
index 945a5dad8c3..29e8fb174f7 100644
--- a/test/e2e/dra/test-driver/app/kubeletplugin.go
+++ b/test/e2e/dra/test-driver/app/kubeletplugin.go
@@ -145,10 +145,11 @@ func (ex *ExamplePlugin) NodePrepareResource(ctx context.Context, req *drapbv1.N
        }
 
        deviceName := "claim-" + req.ClaimUid
-       spec := specs.Spec{
-               // TODO: clarify what version to put here (https://github.com/container-orchestrated-devices/container-device-interface/issues/60).
-               Version: "0.2.0",
-               Kind:    ex.driverName + "/test",
+       vendor := ex.driverName
+       class := "test"
+       spec := &specs.Spec{
+               Version: cdiapi.CurrentVersion,
+               Kind:    vendor + "/" + class,
                // At least one device is required and its entry must have more
                // than just the name.
                Devices: []specs.Device{
@@ -161,12 +162,7 @@ func (ex *ExamplePlugin) NodePrepareResource(ctx context.Context, req *drapbv1.N
                },
        }
        filePath := ex.getYAMLFilePath(req.ClaimUid)
-       cdiSpec, err := cdiapi.NewSpec(&spec, filePath, 1)
-       if err != nil {
-               return nil, fmt.Errorf("create spec: %v", err)
-       }
-
-       buffer, err := yaml.Marshal(*cdiSpec)
+       buffer, err := yaml.Marshal(spec)
        if err != nil {
                return nil, fmt.Errorf("marshal spec: %v", err)
        }
@@ -178,7 +174,7 @@ func (ex *ExamplePlugin) NodePrepareResource(ctx context.Context, req *drapbv1.N
                return nil, fmt.Errorf("failed to write CDI file %v", err)
        }
 
-       dev := cdiSpec.GetDevice(deviceName).GetQualifiedName()
+       dev := cdiapi.QualifiedName(vendor, class, deviceName)
        resp := &drapbv1.NodePrepareResourceResponse{CdiDevices: []string{dev}}
 
        ex.mutex.Lock()
```

pohly:
Something is odd here. `yaml.Marshal` doesn't encode the content of the variable:
```
(dlv) p spec
*github.com/container-orchestrated-devices/container-device-interface/specs-go.Spec {
	Version: "0.5.0",
	Kind: "dra-1856.k8s.io/test",
	Devices: []github.com/container-orchestrated-devices/container-device-interface/specs-go.Device len: 1, cap: 1, [
		(*"github.com/container-orchestrated-devices/container-device-interface/specs-go.Device")(0xc000548700),
	],
	ContainerEdits: github.com/container-orchestrated-devices/container-device-interface/specs-go.ContainerEdits {
		Env: []string len: 0, cap: 0, nil,
		DeviceNodes: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.DeviceNode len: 0, cap: 0, nil,
		Hooks: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.Hook len: 0, cap: 0, nil,
		Mounts: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.Mount len: 0, cap: 0, nil,},}
(dlv) p spec.Devices[0]
github.com/container-orchestrated-devices/container-device-interface/specs-go.Device {
	Name: "claim-b558fe5f-16db-4d95-b17f-c128873a9208",
	ContainerEdits: github.com/container-orchestrated-devices/container-device-interface/specs-go.ContainerEdits {
		Env: []string len: 1, cap: 1, [
			"user_a=b",
		],
		DeviceNodes: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.DeviceNode len: 0, cap: 0, nil,
		Hooks: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.Hook len: 0, cap: 0, nil,
		Mounts: []*github.com/container-orchestrated-devices/container-device-interface/specs-go.Mount len: 0, cap: 0, nil,},}

(dlv) p string(buffer)
"cdiVersion: 0.5.0\ncontainerEdits: {}\ndevices:\n- containerEdits:\n"
```

elezar:
Could you check whether re-exporting the NewSpec method does what you want?
I could review it a bit later tonight and cut a new release with the fix?

pohly:
I found a solution (manual encoding, use version 0.2.0 instead of 0.5.0, use JSON instead of YAML). I filed issues for future investigations.

elezar:
0.2.0 is pretty old. Not sure whether the version is checked anywhere..

As long as you're unblocked we can look into this on Monday.

PR #79, opened 2022-08-22
pkg/cdi: don't abort scan on missing Spec dir.

### LINKED ISSUE #78: device injection fails when /var/run/cdi exists but /etc/cdi does not
Observed with crio 1.24.2. Without /etc/cdi and one file under /var/run/cdi:
```
unresolvable CDI devices example-driver.cdi.k8s.io/test=claim-....
...
212894 newfstatat(AT_FDCWD, "/etc/cdi", 0xc000d2cfa8, AT_SYMLINK_NOFOLLOW) = -1 ENOENT (No such file or directory)
```

With an empty /etc/cdi:
```
219742 newfstatat(AT_FDCWD, "/etc/cdi", {st_mode=S_IFDIR|0755, st_size=4096, ...}, AT_SYMLINK_NOFOLLOW) = 0
219742 openat(AT_FDCWD, "/etc/cdi", O_RDONLY|O_CLOEXEC) = 16
219742 newfstatat(AT_FDCWD, "/var/run/cdi", {st_mode=S_IFDIR|0777, st_size=60, ...}, AT_SYMLINK_NOFOLLOW) = 0
219742 openat(AT_FDCWD, "/var/run/cdi", O_RDONLY|O_CLOEXEC) = 16
219742 newfstatat(AT_FDCWD, "/var/run/cdi/example-driver.cdi.k8s.io-77eac8b8-d949-4004-834d-9da70c7533fa.json", {st_mode=S_IFREG|0644, st_size=151, ...}, AT_SYMLINK_NOFOLLOW) = 0
219742 openat(AT_FDCWD, "/var/run/cdi/example-driver.cdi.k8s.io-77eac8b8-d949-4004-834d-9da70c7533fa.json", O_RDONLY|O_CLOEXEC) = 16
```

klihub:
Ignore scanning errors related to missing Spec directories. This should fix errors where a missing directory (typically /etc/cdi) prevents the resolution of devices using files from subsequent configured directories. Fixes #78 .

### DIFF
--- a/pkg/cdi/spec-dirs.go
+++ b/pkg/cdi/spec-dirs.go
@@ -17,10 +17,10 @@
 package cdi
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 	"path/filepath"
-
-	"github.com/pkg/errors"
 )
 
 const (
@@ -79,6 +79,9 @@ func scanSpecDirs(dirs []string, scanFn scanSpecFunc) error {
 		err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
 			// for initial stat failure Walk calls us with nil info
 			if info == nil {
+				if errors.Is(err, fs.ErrNotExist) {
+					return nil
+				}
 				return err
 			}
 			// first call from Walk is for dir itself, others we skip
--- a/pkg/cdi/spec-dirs_test.go
+++ b/pkg/cdi/spec-dirs_test.go
@@ -186,7 +186,7 @@ devices:
 				classes = map[string]string{}
 			}
 
-			dirs := []string{dir}
+			dirs := []string{"/no-such-dir", dir}
 			err = scanSpecDirs(dirs, func(path string, prio int, spec *Spec, err error) error {
 				name := filepath.Base(path)
 				if err != nil {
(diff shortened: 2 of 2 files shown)


elezar (approved):
Thanks @klihub.

pkg/cdi/spec-dirs.go
```diff
@@ -39,7 +40,7 @@ var (
 	// other Cache instances.
 	DefaultSpecDirs = []string{DefaultStaticDir, DefaultDynamicDir}
 	// ErrStopScan can be returned from a ScanSpecFunc to stop the scan.
-	ErrStopScan = errors.New("stop Spec scan")
+	ErrStopScan = fmt.Errorf("stop Spec scan")
```
bart0sh, line 43: Can you explain why do we need this?

klihub: We don't strictly need it per se. `github.com/pkg/errors` should be deprecated AFAICT. `fmt.Errorf("... %w", err)` can be used to replace `errors.Wrap()` and the stock `errors.Is()` can be used instead of the `pkg/errors` (plus anything else that shows up for error handling in go2).

With that rationale, I did not want to add more usage of `github.com/pkg/errors`. So instead I removed it completely from the file I was touching for this PR.

bart0sh: Sorry, for the confusion. I was trying to say that we can use [New from the errors module](https://pkg.go.dev/errors#New), so we don't need this change.

klihub: Ah, sorry, I forgot about/overlooked that. Much better as we don't really do string formatting here. Updated.

bart0sh (approved):
/lgtm

PR #82, opened 2022-09-05
Fix a read/write data race in the cache.

### LINKED ISSUE #81: there is a data race in CDI version v0.5.0
related PR: https://github.com/containerd/containerd/pull/7331
this PR stuck with a data race when I wanted to update the CDI version to v0.5.0 in contained.
I suspect there is a data race error here but I can't be sure:
1. https://github.com/container-orchestrated-devices/container-device-interface/blob/main/pkg/cdi/cache.go#L449
2. https://github.com/container-orchestrated-devices/container-device-interface/blob/main/pkg/cdi/cache.go#L436

klihub:
When being (re)configured, do not delay dereferencing our fsnotify.Watcher into the start()/watch() goroutine. That is not protected by the lock in Configure() and can trigger a read/write data race. Instead, dereference it while being properly locked and pass it as an argument to watch().

Fixes #81 which should fix [this containerd PR](https://github.com/containerd/containerd/pull/7331) once this fix gets tagged and that containerd PR is updated to use the fix.

This PR also adds a regression test with a gutted version of the containerd test case that triggered this bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 GO_CMD   := go
 GO_BUILD := $(GO_CMD) build
-GO_TEST  := $(GO_CMD) test -v -cover
+GO_TEST  := $(GO_CMD) test -race -v -cover
 
 GO_LINT  := golint -set_exit_status
 GO_FMT   := gofmt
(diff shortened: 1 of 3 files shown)


klihub:
@elezar PTAL. Once the fix is in, you should tag a new version to allow the original [containerd PR](https://github.com/containerd/containerd/pull/7331) to progress.

elezar:
@soulseen could you also take a look at this? I'll review it tomorrow and tag a `v0.5.1`.

soulseen:
@elezar Thanks this looks good.

soulseen (approved):

bart0sh (approved):
/lgtm

PR #83, opened 2022-09-20
pkg/cdi: fix build/vet failures on windows.

klihub:
Fix two build failures for Windows. One in stock code and another in tests, which show up as a 'go vet' failure. Fix a data read/write race which was probably triggered by the latter Windows fix slightly altering timing in some of tests... or just by sheer luck.

Update the github workflow to run sanity checking (lint, vet) both *for* Linux and Windows. The checks always run *on* a Linux distro, but they override GOOS/GOARCH for the toolchain.

Update the github worfklow to do binary test builds/cross-builds both for Linux and Windows.

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -8,6 +8,14 @@ jobs:
   build:
     name: "Run go sanity tools"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - goos: linux
+            goarch: amd64
+          - goos: windows
+            goarch: amd64
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
@@ -21,13 +29,22 @@ jobs:
       run: make fmt
     - name: Vet
       run: make vet
+
+  test:
+    name: "Run tests"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GO_VERSION }}
     - name: Test
       run: make test
 
 
   # Make sure binaries compile on multiple platforms.
   crossbuild:
-    name: Build/Crossbuild Binaries
+    name: "Build / Crossbuild Binaries"
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
@@ -37,7 +54,6 @@ jobs:
             goarch: amd64
           - goos: windows
             goarch: amd64
-
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -1,5 +1,9 @@
 on: [push, pull_request]
 name: Sanity
+
+env:
+  GO_VERSION: '1.17.x'
+
 jobs:
   build:
     name: "Run go sanity tools"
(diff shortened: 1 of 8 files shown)


bart0sh (approved):
/lgtm

pkg/cdi/container-edits_windows.go
```diff
@@ -0,0 +1,25 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cdi
+
+// fillMissingInfo fills in missing mandatory attributes from the host device.
+func (d *DeviceNode) fillMissingInfo() error {
+	return nil
```
elezar, line 24: Should this return an "unimplemented" error instead?

pkg/cdi/cache_test.go
```diff
@@ -770,10 +769,10 @@ devices:
 				for {
 					select {
 					case _ = <-stopCh:
-						go syscall.Sync()
+						osSync()
```
elezar, line 772: Minor question: could this be `go osSync()` instead since go routines themselves are not OS-specific?

.github/workflows/sanity.yml
```diff
@@ -1,14 +1,26 @@
...
   build:
     name: "Run go sanity tools"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - goos: linux
+            goarch: amd64
+          - goos: windows
```
elezar, line 17: I don't suppose `darwin` is of interest to us?

pkg/cdi/cache_test_unix.go
```diff
@@ -0,0 +1,26 @@
+//go:build !windows
+// +build !windows
+
+/*
+   Copyright © 2021 The CDI Authors
```
elezar, line 5: Nit: The date.

elezar (commented):
Some minor comments. I think these are generally no-ops and I'd be happy to merge pending feedback.

(would be nice to update the dates in the license strings).

pkg/cdi/cache_test.go
klihub: Okay. Move goroutine invocation to this end of the function call.

pkg/cdi/cache_test_unix.go
klihub: Argh... does this non-functional change really warrant/require an update of the date ? I already regret including a date there at all. I'd much rather do another PR, dropping all dates from all files and the copyright template, similar to containerd.

pohly: FWIW, the Kubernetes convention seems to be that the date is set once when the file gets added initially and then not updated anymore later on. It's just informative anyway, not legally binding.

pkg/cdi/container-edits_windows.go
klihub: Hmm, good point as such, but I'm not entirely sure since we don't really have a consistent "let all API functions fail on windows" implementation even if we do that.

elezar:
> > Some minor comments. I think these are generally no-ops and I'd be happy to merge pending feedback.
> > (would be nice to update the dates in the license strings).
> 
> My preference would be to file another PR which removes any explicitly spelled out dates from all file headers and the copyright template (and create it if we don't have yet such a thing separately in the repo). This is similar to how, for instance, containerd deals with this. It removes this class of problem once and for all.

That sounds great.

elezar (approved):
Thanks @klihub. 

Sorry about the back and forth on the minor points.

pkg/cdi/cache_test_unix.go
klihub: > FWIW, the Kubernetes convention seems to be that the date is set once when the file gets added initially and then not updated anymore later on. It's just informative anyway, not legally binding.

That sounds like a reasonable approach, too.

PR #85, opened 2022-10-19
Added Annotations to CDI Spec

zvonkok:
Based on the discussion in https://github.com/container-orchestrated-devices/container-device-interface/issues/30 some generated devices like VFs need extra meta information to be handled correctly in virtualized environments.


zvonkok:
For Kata we need to pass through the VF into the container, a VF can have different types and with the BDF the runtime can deduce the right handling. An example CDI spec would look like: 
```
# /etc/cdi/nvidia.yaml
cdiVersion: 0.4.0
kind: nvidia.com/gpu
devices:
- name: gpu0
  annotations:
    bdf: “41:00.0”
    clique-id: “0”
  containerEdits:
    deviceNodes:
    - path: “/dev/vfio/71"```

File: specs-go/config.go [hunk 1]
```diff
@@ -16,8 +16,9 @@ type Spec struct {
 
 // Device is a "Device" a container runtime can add to a container
 type Device struct {
-	Name           string         `json:"name"`
-	ContainerEdits ContainerEdits `json:"containerEdits"`
+	Name           string            `json:"name"`
+	Annotations    map[string]string `json:"annotations,omitempty"`
```
elezar, line 20: I know we haven't really done it up to now, but we could add a docstring for the `Annotations` to explain their intent.

elezar (changes requested):
Thanks @zvonkok. Could you also provide a short description on how these will be used.

Some other points:
* This requires a version bump of the spec. Let us do that now.
* Could we add an example to the readme etc. where this is specified?

specs-go/config.go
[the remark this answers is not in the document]
zvonkok: @klihub I added them "there" because they are not used for any edits of the container. Meta information about the device. Those are annotations per device; if we would need annotations per spec they would need to move to:

```diff
type Spec struct {
	Version string `json:"cdiVersion"`
	Kind    string `json:"kind"`
+       Annotations map[string]string `json:"annotations,omitempty"`
	Devices        []Device       `json:"devices"`
	ContainerEdits ContainerEdits `json:"containerEdits,omitempty"`
}
```
that's why I added them per device: 

```diff
type Device struct {
	Name           string            `json:"name"`
+	Annotations    map[string]string `json:"annotations,omitempty"` // Annotations add meta information per device
	ContainerEdits ContainerEdits    `json:"containerEdits"`
}

```

The annotations here are needed for the runtime not the actual container. If we would need to add annotations to the container, they would need to go to the `ContainerEdits` struct.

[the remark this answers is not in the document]
klihub: Ok, thanks. I missed that it's for consumption by the runtime. Now it makes sense.

zvonkok: Added the docstring, where do you want to have that short description?

zvonkok:
Refactored the README according to md-lint suggestions.

specs-go/config.go
```diff
@@ -16,8 +16,9 @@ type Spec struct {
 
 // Device is a "Device" a container runtime can add to a container
 type Device struct {
-	Name           string         `json:"name"`
-	ContainerEdits ContainerEdits `json:"containerEdits"`
+	Name           string            `json:"name"`
+	Annotations    map[string]string `json:"annotations,omitempty"` // Annotations add meta information per device
```
kad, line 20: Probably good to mention that this metadata is not affecting container metadata.

README.md
```diff
@@ -105,6 +134,32 @@ $ cat > /etc/cdi/vendor.json <<EOF
 }
 EOF
 
+### Using Annotations per device to add meta-information
+
+```bash
+$ mkdir /etc/cdi
+$ cat > /etc/cdi/vendor-annotations.json <<EOF
```
kad, line 141: Probably good to mention that we support both "global" and "per device" annotations, and that these annotations are local to CDI json, meaning not affecting containerEdits parts (which can be separate topic we need to discuss).

kad (commented):
Beside clarification on usage of annotation field, it is ok from my side.

One side note noticed during this PR review, probably good for another PR: I've noticed that in our schema doesn't specify "global" `containerEdits` fields in `schema/schema.json`

specs-go/config.go
elezar: Is there a difference between what is here and:
```
// Annotations add meta information per device. Note these are CDI-specific and do not affect container metadata.       
Annotations map[string]string `json:"annotations,omitempty"` 
```

schema/defs.json
```diff
@@ -23,6 +23,14 @@
         "Env": {
             "$ref": "#/definitions/ArrayOfStrings"
         },
+        "mapStringString": {
```
elezar, line 26: Question: For Kubernetes / Container annotations there are limits to the lenghts of the fields (or at least the name). Do we have any such restrictions that we should add?

specs-go/config.go
kad: My comment was about make it explicit that those annotations are not going to be resulted in editing OCI container spec. So, @elezar suggested edit is good one in my opinion.

zvonkok: Good point.Will update.

README.md
zvonkok: Are we? The annotations are currently only placed at the device level, I wanted to add annotations also at the spec level but was unsure if needed. 

Are you suggesting to add it here as well: 

```diff

// Spec is the base configuration for CDI
type Spec struct {
	Version string `json:"cdiVersion"`
	Kind    string `json:"kind"`
+       Annotations    map[string]string `json:"annotations,omitempty"`
	Devices        []Device       `json:"devices"`
	ContainerEdits ContainerEdits `json:"containerEdits,omitempty"`
}

```

zvonkok: I may have also a  use-case for annotations in the containerEdits, again with Kata, depending on the CDI HW I am exposing I may want to add an containerEdit annotation to trigger a specific behaviour in Kata.

schema/defs.json
zvonkok: In the case of annotations, the key has a limit of 63 characters or less and all the other constraints that are valid for labels too, the values can be arbitrary text  and the sum of all keys + values are not allowed to exceed 256 kB (https://github.com/kubernetes/kubernetes/blob/98742f9d77a57aec44cc05b1daf721973fb029be/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L36). I am not clear if we need a prefix for the per device or if we add per spec annotations that are not changing the Pod spec.

README.md
kad: > Are we? The annotations are currently only placed at the device level, I wanted to add annotations also at the spec level but was unsure if needed.

in `schema/schema.json` I see it added to two places in this PR. In Go structs only in one place.

zvonkok: Ding it, you're right. Let me add it back to the Go struct.

zvonkok: The first thought was to add it globally and locally per device and then added it only to the device. I think globally and per devices makes sense. Yeah let's keep the annotation in the containerEdits for a different discussion.

zvonkok:
@kad @klihub @elezar I added validation of annotations PTAL; I have a general question why are we allowing an empty CDI spec to be valid? What is the purpose? 
For validating annotations, I am at least assuming to have `Kind` and some other fields to be set.

specs-go/config.go
```diff
@@ -3,21 +3,22 @@ package specs
...
 // Spec is the base configuration for CDI
 type Spec struct {
-	Version string `json:"cdiVersion"`
-	Kind    string `json:"kind"`
-
-	Devices        []Device       `json:"devices"`
-	ContainerEdits ContainerEdits `json:"containerEdits,omitempty"`
+	Version        string            `json:"cdiVersion"`
+	Kind           string            `json:"kind"`
+	Annotations    map[string]string `json:"annotations,omitempty"` // Annotations add meta information per device
```
kad, line 12: those are global per file, not per device.

schema/defs.json
zvonkok: @elezar Added the check for annotations PTAL

elezar: Thanks, I'll give it a look. Also note that we're looking at determining the "minimum spec version" in #93 so that producers can fill the information that they need and then save the spec with the minimum version. We would need to add a function for checking whether a `v0.6.0` spec is required in the case of annotations. If this gets merged before then, I can add that though.

schema/schema.go
```diff
@@ -187,8 +190,76 @@ func (s *Schema) ValidateData(data []byte) error {
 			return errors.Wrap(err, "failed to JSON remarshal data for validation")
 		}
 	}
+	if err = s.validate(schema.NewBytesLoader(data)); err != nil {
+		return err
+	}
+	return s.validateSpecAnnotations(data)
```
elezar, line 196: I would personally insert this before the call to `s.validate`. I would also in this case unconditionally call `yaml.Unmarshal` and pass `any` to the annotation validation methods below.

schema/schema.go
```diff
@@ -187,8 +190,76 @@ func (s *Schema) ValidateData(data []byte) error {
 			return errors.Wrap(err, "failed to JSON remarshal data for validation")
 		}
 	}
+	if err = s.validate(schema.NewBytesLoader(data)); err != nil {
```
elezar, line 193: Just a note (@klihub to confirm): I don't think that this validation is enabled by default, meaning that a producer will have to explicitly opt-in to validation here. Is there something that can be done when reading / writing a spec instead?

schema/schema_test.go
```diff
@@ -247,7 +248,7 @@ func scanAndValidate(t *testing.T, scm *schema.Schema, dir string, isValid bool,
 			if name := info.Name(); filepath.Ext(name) != ".json" || name == "empty.json" {
 				return nil
 			}
-			//fmt.Printf("*** processing %s...\n", path)
+			fmt.Printf("*** processing %s...\n", path)
```
elezar, line 251: Should this change be here?

specs-go/config.go
elezar: Does it make sense to have the docstring preceding the field, or is a trailing comment also common practice?

schema/schema.go
```diff
@@ -187,8 +190,76 @@ func (s *Schema) ValidateData(data []byte) error {
...
+		return errors.Wrap(err, "failed to parse spec annotations data for validation")
+	}
+	return validateAnnotations(annotationsIfc)
+}
+
+// validateAnnotations according to Kubernetes rules
+func validateAnnotations(annotationsIfc map[string]interface{}) error {
+	annotations := make(map[string]string)
+	for key, value := range annotationsIfc {
+		annotations[key] = fmt.Sprintf("%v", value)
```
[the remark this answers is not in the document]
zvonkok: Nope, good catch, updated.

schema/schema.go
zvonkok: Why before? I want to have a valid document before I go and check the annotations.

specs-go/config.go
elezar: Sorry, missed this one. I think the readme where we show the example we could expand on this. I don't think it's a blocker for the PR though.

schema/schema.go
elezar: Sorry, I mean, there is a conditional `yaml.Unmarshal` call to heck whether we can re-encode the YAML as JSON if required. I was suggesting we reuse this unmarshalled object for our annotation validations instead of re-Unmarshalling the data there.

zvonkok: Ah right, thanks for the point!! PTAL

schema/schema.go
zvonkok: @klihub Any comments on ^^ Thanks!

schema/schema_test.go
zvonkok: Reverted.

schema/schema.go
klihub: Argh, sorry guys for not responding/missing this one...

Yes, that validation is not in effect/performed by default. You have to enable it by something like
```
import (
    "flag"
...
    "github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
    "github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/validate"
    "github.com/container-orchestrated-devices/container-device-interface/schema"
...
)

var (
    schemaName string
)

// Set up JSON schema based Spec validation.
func setValidationSchema() error {
    s, err := schema.Load(schemaName)
    if err != nil {
        return fmt.Errorf("failed to load JSON schema %s: %w", schemaName, err)        
    }
    cdi.SetSpecValidator(validate.WithSchema(s))
    return nil
}

func main() {
    flag.StringVar(&schemaName, "schema", schema.BuiltinSchemaName,
        "validation schema file or URI, use " + schema.NoneSchemaName + "for no validation")
    ...
    flag.Parse()
    ...
    setValidationSchema()
    ...
}
```

There is a more accurate working example of that in `cmd/cdi/cmd/root.go` IIRC... 

The reason to not do this by default is to keey by default the extremely problematic `github.com/xeipuuv/gosjon*` packages out from the vendored dependencies of our downstream projects. This is good enough for most, but not all. For instance Kubernetes refuses to have those packages even in their `go.sum` (and therefore ATM importing this package).  Therefore we have #97 open...

klihub:
> I have a general question why are we allowing an empty CDI spec to be valid? What is the purpose?

AFAICT, it is partly for historical reasons. IIRC how it came to be, Renaud's initial code was explicitly allowing for it, so we kept that functionality. My guess is that the orignal idea was to let vendor-specific utilities run at boot-time and generate vendor-specific 'static' Spec files in /etc (maybe by scanning /dev, and/or doing udev enumeration) and allow any empty files to indicate that the utilities did run (maybe using comments to that effect in the generated Spec files) but not trigger errors in the actual cdi Spec-handling packages. But I'm just guessing here...

schema/schema.go
elezar: @zvonkok I have created https://github.com/zvonkok/container-device-interface/pull/1 which adds basic validation of annotations for devices and specs where other validation is currently performed. I think this circumvents the whole schema validation issue that we're looking at currently.

I believe that since we use `ReadSpec` to load the spec (which invokes `newSpec` where `spec.validate` is called), this should also validate the annotations when loading from disk.

elezar (commented):
@zvonkok I have reworked this PR so that it can be merged into the current `main` branch. This includes checks for the minimum version, for example.

I have also reworked the annotation validation to reduce the imported `k8s` dependencies.

@klihub @bart0sh @kad since I've also updated this PR please take a look now.

bart0sh:
It would be nice to cover new code with unit tests.
Other than that lgtm.

klihub:
This needs a rebase and conflict resolution. Other than that, I think it looks fine.

elezar:
> This needs a rebase and conflict resolution. Other than that, I think it looks fine.

Done.

internal/validation/validate.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/container-orchestrated-devices/container-device-interface/internal/validation/k8s"
+)
+
+// ValidateSpecAnnotations checks whether spec annotations are valid.
+func ValidateSpecAnnotations(name string, any interface{}) error {
```
elezar, line 27: @klihub I believe you had questions / concerns about the validation of the annotations.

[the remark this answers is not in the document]
elezar: This function is called from the schema validation, where all we have is an `interface{}` type. This handles the conversion of this type to a `map[string]string`. Since this is required for two separate parts of the schema / spec (top-level annotations and device-level annotation), I added this helper here.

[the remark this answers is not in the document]
elezar: Maybe I can rename this helper to `ValidateSpecAnnotationSchema` instead.

Antother option is to drop the schema validation entirely which would clean up this PR significantly.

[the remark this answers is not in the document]
klihub: Yes, it is sort of tempting... Then again, if at all possible it is better to catch Spec errors early (when seen first time by the CDI package) than late (when an injection goes wrong in the runtime). So I guess then some basic annotation verification would be a good idea.

That said, I don't know if these checks are really that helpful in that regard. AFAICT we don't plan to inject any of these annotations to a container and we are doing K8s pod/container annotations here. If we don't inject them, some entity somewhere along the runtime-runc/crun/kata path will do some semantic interpretation of the annotated data... which inevitable leaves the door open for delayed errors.

zvonkok:
@kad @klihub @bart0sh Anything missing to merge this?

klihub:
@zvonkok Not from my point of view. As I stated in my last review comment I'm fine with merging this as such, or if @elezar wants to drop `ValidateSpecAnnotations`, I'm also fine with that.

elezar:
> @zvonkok Not from my point of view. As I stated in my last review comment I'm fine with merging this as such, or if @elezar wants to drop `ValidateSpecAnnotations`, I'm also fine with that.

Since we don't validate by default, I think we can leave it in for the time being. If you're good with it @klihub let's get this merged and I we can bump a new version.

internal/validation/k8s/objectmeta.go
```diff
@@ -0,0 +1,57 @@
...
+	if err := ValidateAnnotationsSize(annotations); err != nil {
+		errors = multierror.Append(errors, fmt.Errorf("%v is too long: %v", path, err))
+	}
+	return errors
+}
+
+// ValidateAnnotationsSize validates that a set of annotations is not too large.
+func ValidateAnnotationsSize(annotations map[string]string) error {
+	var totalSize int64
+	for k, v := range annotations {
```
bart0sh, line 50: Would it make sense to get rid of this second loop? It can be done by either moving size validation code inside a first loop or by calling ValidateAnnotationSize from inside the first loop.

internal/validation/validate.go
```diff
@@ -0,0 +1,56 @@
...
+	case map[string]interface{}:
+		annotations := make(map[string]string)
+		for k, v := range v {
+			if s, ok := v.(string); ok {
+				annotations[k] = s
+			} else {
+				return fmt.Errorf("invalid annotation %v.%v; %v is not a string", name, k, any)
+			}
+		}
+		return validateSpecAnnotations(name, annotations)
```
bart0sh, line 42: There is also double looping here. Not sure it's feasible to fix it as well, just wanted to point out.

internal/validation/k8s/objectmeta.go
elezar: This code is taken from the `k8s` validation code. The only changes I made was to remove code that was not needed for the validation. I am happy to revisit if we think we should implement our own logic.

internal/validation/validate.go
elezar: You mean looping twice? The first loop performs a type conversion so that the actual validation can be applied. 

Do you have a suggestion on how this can be achieved more cleanly?

pkg/cdi/version.go
```diff
@@ -115,6 +117,23 @@ func (r requiredVersionMap) requiredVersion(spec *cdi.Spec) version {
...
+func requiresV060(spec *cdi.Spec) bool {
+	// The v0.6.0 spec allows annotations to be specified at a spec level
+	for range spec.Annotations {
+		return true
+	}
+
+	// The v0.6.0 spec allows annotations to be specified at a device level
+	for _, d := range spec.Devices {
+		for range d.Annotations {
+			return true
```
bart0sh, line 130: Do we need a loop here if we unconditionally return on the first iteration?

internal/validation/k8s/objectmeta.go
bart0sh: It's up to you. I don't think we can gain a lot by changing it. You can consider this as a nit.

internal/validation/validate.go
bart0sh: my bad. I thought first loop is also over the annotations.

klihub (approved):
LGTM.

pkg/cdi/version.go
elezar: The intent of the loop is to detect whether a `v0.6.0` spec is required. This is required if ANY annotions are detected. 

We could use:
```
	for _, d := range spec.Devices {
   		if len(d.Annotations) > 0 {
			return true
```
instead, but I thought the loop was cleaner than the conditional in this case.

bart0sh: Not for me, but I'm ok with it.

Issue #88, opened 2022-11-01
dependencies: remove github.com/hashicorp/go-multierror and github.com/hashicorp/errwrap


pohly:
Using CDI in Kubernetes adds two new dependencies:

```
$ go mod why -m github.com/hashicorp/go-multierror
# github.com/hashicorp/go-multierror
k8s.io/kubernetes/pkg/kubelet/cm/dra
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi
github.com/hashicorp/go-multierror

$ go mod why -m github.com/hashicorp/errwrap
# github.com/hashicorp/errwrap
k8s.io/kubernetes/pkg/kubelet/cm/dra
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi
github.com/hashicorp/go-multierror
github.com/hashicorp/errwrap
```

Can those two dependencies be removed? Go natively has support for wrapping errors and combining multiple multiple errors into one can be implemented with custom code instead of dragging in a dependency.

PR #89, opened 2022-11-01
remove usage of github.com/hashicorp/go-multierror

pohly:
The internal replacement is a simpler implementation which only supports enough functionality for the places where github.com/hashicorp/go-multierror was used before.

Not depending on github.com/hashicorp/go-multierror and thus github.com/hashicorp/errwrap makes it easier to use CDI in Kubernetes because we don't need to add those other dependencies.

Fixes: #88

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.17
 
 require (
 	github.com/fsnotify/fsnotify v1.5.1
-	github.com/hashicorp/go-multierror v1.1.1
 	github.com/opencontainers/runc v1.1.2
 	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
 	github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e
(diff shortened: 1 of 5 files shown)


kad:
@pohly the dependency seems to be still present as indirect, is that expected?

Second question, is `multierror.test` binary really needed in the tree?

pohly:
> @pohly the dependency seems to be still present as indirect, is that expected?

Yes:

```
# github.com/hashicorp/go-multierror
github.com/container-orchestrated-devices/container-device-interface/cmd/cdi/cmd
github.com/opencontainers/runtime-tools/generate
github.com/opencontainers/runtime-tools/validate
github.com/hashicorp/go-multierror
```

github.com/opencontainers/runtime-tools would have to be modified to get rid of it completely. May be useful, but doesn't affect us that much because it doesn't get pulled into Kubernetes (I checked).

> Second question, is `multierror.test` binary really needed in the tree?

Oops, no. Looks like I was relying on Go projects filtering out "*.test" during `git add` and this project doesn't do that. Fixed.

kad:
ok. if that is not propagating to k8s as import, then it is lgtm from me.

elezar (approved):
Thanks @pohly

PR #94, opened 2022-11-08
Fix RemoveSpec, WriteSpec example usage in doc

byako:
RemoveSpec and WriteSpec appear to be in SpecDB, not in Registry, hence this PR to make doc correspond to code.


klihub:
Closing as duplicate of #96, choosing this one as the victim mostly because this had the bad fortune to trigger during the tests a read-write data race which should now be fixed by #98 .

PR #95, opened 2022-11-09
Use current fsnotify library, drop viper dependency

liggitt:
Working through the review of integrating this into kubernetes in https://github.com/kubernetes/kubernetes/pull/111023, I found some dependency issues:

* cmd is using a different fsnotify library than pkg, pinned to an obsolete version (v1.4.7); this PR collapses cmd onto the same fsnotify dependency as pkg
* cmd is using an old version of cobra that transitively pulls in viper (https://github.com/spf13/cobra/issues/1597); this PR updates to cobra v1.6.0 which drops the viper transitive dependency (go.sum gets *way* cleaner)


klihub (approved):
LGTM.

liggitt:
I don't have context to know what the maintainers want to do with the cobra dep. Multi-module repos can be painful to manage, so I wouldn't assume folks here would want to do that.

I'd probably recommend merging this as a first easy step, then letting a module carve-out effort be done as a follow-up if desired

PR #96, opened 2022-11-09
fix cdi doc missing SpecDB()

moshe010:
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/pkg/cdi/doc.go
+++ b/pkg/cdi/doc.go
@@ -196,7 +196,7 @@
 //         return fmt.Errorf("failed to generate Spec name: %w", err)
 //     }
 //
-//     return registry.WriteSpec(spec, specName)
+//     return registry.SpecDB().WriteSpec(spec, specName)
 // }
 //
 // Similary, generating and later cleaning up transient Spec files can be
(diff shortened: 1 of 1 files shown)


klihub:
LGTM.

That said, I again would like to get rid of the artificial extra registry interfaces (all Registry[A-Z]*). They are neither useful in any way nor correspond to anything in the underlying implementation as all of them are directly implemented by the underlying Cache instance.

Moreover, before ever considering tagging a stable version, I'd like to go back to the more golang-ish very first implementation. There the registry was an implicit/unexported default `Cache` instance and its full interface was available as package-global functions, operating on that implicit default instance (just like for instance `flag` is doing with `CommandLine`). Then operating on the registry would be simply `cdiXyzzyFoobar()` instead of needlessly having to go `cdi.GetRegistry().XyzzyFoobar()`. WDYT, folks ?

klihub (approved):

bart0sh:
@klihub +1 Getting rid of nested interfaces sounds good to me.

bart0sh (approved):
/lgtm

PR #98, opened 2022-11-11
pkg/cdi: fix GetSpecErrors()/refresh read/write data race.

klihub:
Avoid a read/write data race between refresh() and GetSpecErrors() by taking the cache lock during GetSpecErrors(). While at it, also make a copy and return that instead of the original error slice in the Spec.

Signed-off-by: Krisztian Litkey <krisztian.litkey@intel.com>


elezar (approved):
Thanks @klihub

bart0sh (approved):
/lgtm

PR #99, opened 2022-11-11
cmd, pkg, schema: stop using github.com/pkg/errors.

klihub:
Get rid of `github.com/pkg/errors` dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/opencontainers/runc v1.1.2
 	github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb
 	github.com/opencontainers/runtime-tools v0.9.1-0.20221107090550-2e043c6bd626
-	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.6.0
 	github.com/stretchr/testify v1.7.0
 	github.com/xeipuuv/gojsonschema v1.2.0
--- a/pkg/cdi/annotations.go
+++ b/pkg/cdi/annotations.go
@@ -17,9 +17,9 @@
 package cdi
 
 import (
+	"errors"
+	"fmt"
 	"strings"
-
-	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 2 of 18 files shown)


go.mod
```diff
@@ -1,13 +1,12 @@
 module github.com/container-orchestrated-devices/container-device-interface
 
-go 1.17
+go 1.19
```
elezar, line 3: Question: How does this affect consumers?

pkg/cdi/annotations.go
```diff
@@ -17,9 +17,9 @@
 package cdi
 
 import (
+	"errors"
```
elezar, line 20: Should we use `fmt.Errorf` over `errors.New` for all errors even if no formatting is performed?

elezar (commented):
One question about the Golang version bump.

go.mod
klihub: Well, it forces them to use the same version. And that's a good point you make. I assumed all of our major downstream projects have switched to go 1.19 by now, but now that I checked it's only kubernetes and cri-o. containerd is still at 1.18. So I guess we need to switch to that instead until containerd makes the bump.

pkg/cdi/annotations.go
klihub: Would there be some real benefit to it ? This `errors` is golang stdlib (so always present/available by the toolchain) and we anyway depend on it because we use `errors.Is()` elsewhere in the code.

elezar: Was more of a general question. Not a blocker.

mythi: a related comment is to make the cdi errors hierarchy so that, e.g., `goerr113` linter is happy. In particular,
"`error` should not be created dynamically from scratch but by the wrapping the static (package-level) error."

```
$ golangci-lint run --disable-all -E goerr113
```

go.mod
elezar: In the `container-toolkit` (where we also consume this) we currently use `1.17`. I don't think we're blocked on increasing this though.

go.mod
```diff
@@ -1,13 +1,12 @@
 module github.com/container-orchestrated-devices/container-device-interface
 
-go 1.17
+go 1.18
```
elezar, line 3: Note that podman uses 1.17. Is this a hard requirement?

klihub: No, not a hard requirement, I guess. But IMO there is no reason to stay on 1.17, especially since it has been [EOL'd](https://endoflife.date/go) more than 3 months ago. 1.18 has been released more than 6 months ago and also 1.19 more than 3 months ago.

elezar: OK, that's fair. I have created https://github.com/containers/podman/pull/16530 to update the version in podman.

klihub: Yes, definitely we can. I can switch it back to 1.17 and we can bump it later, preferably directly to 1.19 if all of our downstream/potential downstream has switched.

klihub: Pushed again with the go version bump patch dropped.

elezar: Thanks. I'll follow up on the podman PR next week and we can revisit once that's sorted out.

pkg/cdi/annotations.go
elezar: @mythi I may have been a bit hasty in resolving this thread. I have created #102 to address the lint errors. I think that this is out of scope of this PR and also doesn't make the situation worse than what it currently is.

elezar (approved):
Thanks @klihub. After dropping the golang version bump this looks good.

pkg/cdi/annotations.go
mythi: @elezar no worries, I think it's fine like this and we can work on the improvements through #102

PR #87, opened 2022-10-27
dependencies: update opencontainers packages

pohly:
The update runtime-tools contain
https://github.com/opencontainers/runtime-tools/pull/755 and thus CDI will have less dependencies when only used to write files without validation.

I have verified that this removes github.com/blang/semver and github.com/xeipuuv from Kubernetes,
Fixes: https://github.com/container-orchestrated-devices/container-device-interface/issues/80

PR #100, opened 2022-11-11
dependencies: update opencontainers packages (rebased #87)

klihub:
The update runtime-tools contain
https://github.com/opencontainers/runtime-tools/pull/755 and thus CDI will have less dependencies when only used to write files without validation.

Signed-off-by: Patrick Ohly <patrick.ohly@intel.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,8 @@ go 1.17
 require (
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/opencontainers/runc v1.1.2
-	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
-	github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e
+	github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb
+	github.com/opencontainers/runtime-tools v0.9.1-0.20221107090550-2e043c6bd626
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.6.0
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -5,8 +5,8 @@ go 1.17
 require (
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/opencontainers/runc v1.1.2
-	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
-	github.com/opencontainers/runtime-tools v0.0.0-20190417131837-cd1349b7c47e
+	github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb
```
elezar, line 8: Question: I don't see `v1.0.3` or `v0.9.1` tags in these repos. Is this because these are pseudo versions that gomod bumps to the next patch release?

go.sum
```diff
@@ -1,6 +1,6 @@
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
-github.com/blang/semver v3.5.1+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
+github.com/blang/semver/v4 v4.0.0 h1:1PFHFE6yCCTv8C1TeyNNarDzntLi7wMI5i/pzqYIsAM=
```
elezar, line 2: Is this now a transitive dependency? Did we run `go mod tidy` to clean things up?

elezar (approved):
Some minor questions / comments, but not blockers.

go.mod
klihub: Runtime tools has been going now for years untagged. We have to hand-pick/lock its version at or past the two commits that prevent the xeipuuv/jsonschema dependencies from always getting pulled in downstream. That sha1 is the tip of that tree. Same story for runtime-spec wrt. tagging. Last tagged one (1.0.2) is from January 2020. We don't need the absolute latest version, but we need newer than 1.0.2.

bart0sh (approved):
/lgtm

PR #110, opened 2023-02-20
Add cdi.WriteSpec function to write specs to file

elezar:
For producers that do not interact with a CDI registry / cache having access to a single method for writing a generated CDI specification to file is useful. The alternative is to expose validation and in-place updating of spec files through the API.

### DIFF
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -114,9 +114,10 @@ func newSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
 	return spec, nil
 }
 
-// Write the CDI Spec to the file associated with it during instantiation
-// by newSpec() or ReadSpec().
-func (s *Spec) write(overwrite bool) error {
+// WriteSpec writes the specified raw CDI spec to the specified path.
+// An overwrite flag controls whether an existing file is overwritten or if an
+// error should be returned.
+func WriteSpec(raw *cdi.Spec, path string, overwrite bool) error {
 	var (
 		data []byte
 		dir  string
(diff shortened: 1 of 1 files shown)


bart0sh:
Not that I'm against this, but it could be confusing to have 2 APIs for writing specs. AFAIR [this commit](https://github.com/container-orchestrated-devices/container-device-interface/commit/98524c3d232a31f6cf6b326cc3023aee743e5acb) was addressing similar confusion.

elezar:
> Not that I'm against this, but it could be confusing to have 2 APIs for writing specs. AFAIR [this commit](https://github.com/container-orchestrated-devices/container-device-interface/commit/98524c3d232a31f6cf6b326cc3023aee743e5acb) was addressing similar confusion.

One of the issues with the linked change (and the existing API) is that it does not allow me as a CDI spec producer to write a spec to an arbitrary path. I suppose there are ways around this -- for example creating a `Registry` / `Cache` with a path set to the desired output path, but this seems a little roundabout to me.

Considering my requirements from the point of view of the NVIDIA CDI generation tooling:

1. I can validate an in-memory spec without writing a spec to file
2. I can output (and validate) a spec to an `io.Writer` so that I can also output it to `STDOUT`
3. I can write (and validate) a spec to a specific path and have the correct format determined from the file path.

This PR doesn't really cover these use cases and would need additional work. This is why it is marked as a draft.

If I recall correctly, we did have some discussion around separating the producer and consumer use cases, with the former not specifically needing the functionality provided by the registry (unless we are looking for global consistency across all specs in the target folders).

PR #116, opened 2023-03-15
Create dependabot.yml

elezar:
This change adds a dependabot file for the repo. 

This is taken from the one used in podman: https://github.com/containers/podman/blob/68bf49799d6085bcc0752e15630ad18ff4d99e19/.github/dependabot.yml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: weekly
+  labels:
+    - "release-note-none"
+  open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


klihub (approved):
LGTM. I wouldn't necessarily run it daily, but that's fine, too.

elezar:
> LGTM. I wouldn't necessarily run it daily, but that's fine, too.

Can increase to `weekly` instead.

klihub:
> > LGTM. I wouldn't necessarily run it daily, but that's fine, too.
> 
> Can increase to `weekly` instead.

Yeah. Maybe it would be better, to see what kind of PR-traffic we get.

klihub (approved):

PR #125, opened 2023-03-15
Removed release-note-none label from dependabot

elezar:
The dependabot PRs created fail due to a missing label.

This PR removes the setting.

An alternative would be to add the label to the project.

See for example: https://github.com/container-orchestrated-devices/container-device-interface/pull/124#issuecomment-1469927721


klihub (approved):
LGTM.

PR #126, opened 2023-03-15
Prepend yaml separator (---) to yaml output

### LINKED ISSUE #111: Include yaml separator (---) when outputing specs in yaml mode
We could use something like:
```
	if format == formatYAML {
		_, err := output.Write([]byte("---\n"))
		if err != nil {
			return fmt.Errorf("failed to write YAML separator: %v", err)
		}
	}
	_, err := output.Write(data)
	if err != nil {
		return fmt.Errorf("failed to write data: %v", err)
	}

	return nil
```

This makes concatenating multiple YAML files simpler.

elezar:
This includes a YAML separator (`---`) at the start of the YAML output which makes working with multiple spec files simpler.

Closes #111

### DIFF
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -132,6 +132,7 @@ func (s *Spec) write(overwrite bool) error {
 
 	if filepath.Ext(s.path) == ".yaml" {
 		data, err = yaml.Marshal(s.Spec)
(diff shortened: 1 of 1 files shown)


klihub (approved):
LGTM

bart0sh (approved):
/lgtm

PR #137, opened 2023-04-24
Support dots in Kind name segment

adrianchiris:
CDI spec states that dots ('.') are allowed.
Add support in parser for this.

reference: https://github.com/container-orchestrated-devices/container-device-interface/blob/main/SPEC.md#kind

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -27,6 +27,7 @@ Released versions of the spec are available as Git tags.
 | v0.4.0 |   | Added `type` field to Mount specification |
 | v0.5.0 |   | Add `HostPath` to `DeviceNodes` |
 | v0.6.0 |   | Add `Annotations` field to `Spec` and `Device` specifications |
+|            |    | Allow dots (`.`)  in name segment of `Kind` field |
 
 *Note*: The initial release of a **spec** with version `v0.x.0` will be tagged as
 `v0.x.0` with subsequent changes to the API applicable to this version tagged as `v0.x.y`.
--- a/pkg/cdi/qualified-device.go
+++ b/pkg/cdi/qualified-device.go
@@ -25,13 +25,9 @@ import (
 //
 //	"<vendor>/<class>=<name>".
 //
-// A valid vendor name may contain the following runes:
+// A valid vendor and class name may contain the following runes:
 //
 //	'A'-'Z', 'a'-'z', '0'-'9', '.', '-', '_'.
-//
-// A valid class name may contain the following runes:
-//
-//	'A'-'Z', 'a'-'z', '0'-'9', '-', '_'.
 //
 // A valid device name may containe the following runes:
 //
@@ -98,7 +94,7 @@ func ValidateVendorName(vendor string) error {
 // A class name may contain the following ASCII characters:
 //   - upper- and lowercase letters ('A'-'Z', 'a'-'z')
 //   - digits ('0'-'9')
-//   - underscore and dash ('_', '-')
+//   - underscore, dash, and dot ('_', '-', and '.')
 //
 // Deprecated: use parser.ValidateClassName instead
 func ValidateClassName(class string) error {
--- a/pkg/cdi/qualified-device_test.go
+++ b/pkg/cdi/qualified-device_test.go
@@ -55,6 +55,13 @@ func TestQualifiedName(t *testing.T) {
 			isQualified: true,
 		},
 		{
+			device:      "vendor1.com/class.subclass=dev1",
+			vendor:      "vendor1.com",
+			class:       "class.subclass",
+			name:        "dev1",
+			isQualified: true,
+		},
+		{
 			device:      "other-vendor1.com/class_1=dev_1",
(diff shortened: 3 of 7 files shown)


elezar (commented):
Thanks for the contribution @adrianchiris. I don't think I have an issue with this.

@klihub can you recall why dots were not included in device classes when this was initially implemented?

One thing that would also have to be added is that we would have to add a check and an associated test to the `requiresV060` function in `pkg/cdi/version.go`. This would mean that a lower spec version can be generated if a non-dotted class name is used to ensure better compatibility with our clients.

adrianchiris:
Ack, thx for the feedback will work on adding the needed check & test

adrianchiris:
> One thing that would also have to be added is that we would have to add a check and an associated test to the requiresV060 function in pkg/cdi/version.go. This would mean that a lower spec version can be generated if a non-dotted class name is used to ensure better compatibility with our clients.


Done.

elezar:
lgtm

@klihub @bart0sh any comments here?

klihub:
> Thanks for the contribution @adrianchiris. I don't think I have an issue with this.
> 
> @klihub can you recall why dots were not included in device classes when this was initially implemented?

I think it was simply considered unnecessary (there was no immediate need for it) and cleaner to only accept dots in the vendor part where it was clearly needed (to allow for DNS-like vendor prefixes).

klihub:
> lgtm
> 
> @klihub @bart0sh any comments here?

I have no strong feelings. I don't see why dots would be necessary or useful in the class part. But you guys have been playing around much more with generating Specs and if you say it is necessary/helpful, then I believe you.

> CDI spec states that dots ('.') are allowed.
> Add support in parser for this.

A much simpler alternative would have been to simply update the specs reflect the implementation, stating that dots are only allowed in the vendor prefix. But as I said, no strong feelings and I'm fine with this.

bart0sh (approved):
I'm ok with this.

adrianchiris:
Some context on what triggered me to add this:

im working on adding support via DRA to network resources for NVIDIA NICs
part of this work is to generate CDI files for container runtime to consume.

there may be different network device types (e.g SRIOV virtual function[1] and subfunctions[2])

So, i was thinking on how to name them in CDI.

as spec allowed dots in name segment this is what I had in mind for the kind string (and its what triggered this PR):
```
nvidia.com/net.sf=<SF device name>
nvidia.com/net.vf=<VF device address>
```
since the first part usually refers to the vendor (at least that what i got from the examples within CDI proj)
it made sense to me to chose the kind it as above.

some alternatives that conform to current implementation (without dot in name segment)

```
nvidia.com/net-sf=<SF device name>
nvidia.com/net-vf=<VF device address>
```

```
net.nvidia.com/sf=<SF device name>
net.nvidia.com/vf=<VF device address>
```

[1] https://en.wikipedia.org/wiki/Single-root_input/output_virtualization
[2] https://docs.kernel.org/networking/devlink/devlink-port.html#subfunction

adrianchiris:
> I discussed this briefly in the COD Working group meeting, and have no objections. This also aligns with what is allowed for [Kubernetes Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) and does provide more flexibility from a user perspective.

Thanks for bringing this one up for discussion.


> @adrianchiris as a last update, please add a line to the "Released versions of the spec are available as Git tags." table in SPEC.md that indicates this relaxation. (I think we're missing one for the relaxation of the device name too which I will add in a follow-up).

Done. updated v0.6.0 entry. LMK if wording is OK

bart0sh:
@elezar Isn't it a backward-incompatible change? If it is we should consider incrementing spec version.

elezar:
> @elezar Isn't it a backward-incompatible change? If it is we should consider incrementing spec version.

It is backward incompatible, but we haven't released (tagged) the `v0.6.0` spec yet. Meaning that this change would be included as part of the `v0.6.0` spec once we tag that.

SPEC.md
```diff
@@ -26,7 +26,7 @@ Released versions of the spec are available as Git tags.
 | v0.3.0 |   | Initial tagged release of Spec |
 | v0.4.0 |   | Added `type` field to Mount specification |
 | v0.5.0 |   | Add `HostPath` to `DeviceNodes` |
-| v0.6.0 |   | Add `Annotations` field to `Spec` and `Device` specifications |
+| v0.6.0 |   | Add `Annotations` field to `Spec` and `Device` specifications. Allow dots (`.`)  in name segment of `Kind` field |
```
elezar, line 29:
```suggestion
| v0.6.0 |   | Add `Annotations` field to `Spec` and `Device` specifications |
|            |    | Allow dots (`.`)  in name segment of `Kind` field |
```

elezar (approved):
Thanks @adrianchiris.

This looks good now.

adrianchiris:
@bart0sh @elezar we OK with merging this one ?

klihub (approved):

PR #138, opened 2023-05-02
Ensure that spec files are created world readable

elezar:
In order to allow non-root applications to read the generated specs, the files must be created with 644 permissions instead of the 600 permissions that are the default for os.CreateTemp.

This change calls os.Chmod on the generated temporary file before renaming it to the final filename.

### DIFF
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -156,6 +156,11 @@ func (s *Spec) write(overwrite bool) error {
 		return fmt.Errorf("failed to write Spec file: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


elezar:
@klihub @kad mentioned that we may not want to make this the default behaviour -- especially in the context of device plugins. I will revisit this -- suplying some option when constructing the cache or the spec instead of always setting this to `644`.

I have created https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/381 to do this in our client, so this is not a blocker from our side.

PR #139, opened 2023-05-04
codespell: workflow, config, typos fixed

yarikoptic:
While listening to @elezar  giving a presentation, decided to check the CDI out: Having typos free code would be very useful for 1.0 release ;)

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,*.pdf,*.svg
+#
+# ignore-words-list =
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,4 +1,4 @@
 [codespell]
-skip = .git,*.pdf,*.svg
+skip = .git,*.pdf,*.svg,*.sum,*.mod
 #
 # ignore-words-list =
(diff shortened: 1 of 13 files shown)


elezar (approved):
Thanks @yarikoptic. lgtm

@klihub @bart0sh any objections?

.codespellrc
```diff
@@ -0,0 +1,4 @@
+[codespell]
+skip = .git,*.pdf,*.svg
+# BU - part of JP3t17pCcGlemwknint6hfoeCVQrEMVwxRLRjXpq+BU=
```
elezar, line 3: @yarikoptic as someone who is not too familiar with codespell what does this represent?

yarikoptic:
merged `main`, resolved conflicts

.codespellrc
elezar: This string is in a `go.sum` file. Does it not make more sense to add `*.sum` to the `skip` list?

elezar (changes requested):
Thanks for the explaination. Please rather add `go.sum` and `go.mod` to the skip list instead of handling the checksum as a word.

.codespellrc
yarikoptic: indeed, done!

elezar (approved):
lgtm

PR #149, opened 2023-08-04
Rename test package to parser_test

elezar:
This change renames the package for parser tests to parser_test to further limit the dependencies that are required by downstream consumers.

See the discussion here: https://github.com/kubernetes-sigs/kind/pull/3290#discussion_r1252000400

### DIFF
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -14,10 +14,12 @@
    limitations under the License.
 */
 
-package parser
+package parser_test
 
 import (
 	"testing"
+
+	"github.com/container-orchestrated-devices/container-device-interface/pkg/parser"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


bart0sh:
@elezar Did you test if it actually works? I thought that `go get` doesn't add test dependencies by default. It has  optional `-t` command line switch to download test dependencies, but it's optional.

klihub:
LGTM. I think whenever possible/reasonable (IOW, unless there is a lot of complex unexported code that warrants dedicated unit testing), it is a good idea to put tests into it's own package namespace. This happens automatically if/when ginkgo/gomega is used.

pohly:
> it is a good idea to put tests into it's own package namespace. This happens automatically if/when ginkgo/gomega is used.

Is that so? Nothing prevents someone from writing a `foo_test.go` with `package foo` where ginkgo/gomega are used to test unexported symbols.

klihub:
> > it is a good idea to put tests into it's own package namespace. This happens automatically if/when ginkgo/gomega is used.
> 
> Is that so? Nothing prevents someone from writing a `foo_test.go` with `package foo` where ginkgo/gomega are used to test unexported symbols.

I've always `ginkgo bootstrap`ped test suites (without extra options) and then you end up with ${pkg}_test as the package for the test suite. Now that I checked the docs indeed state, that "Ginkgo defaults to setting up the suite as a *_test package to encourage you to only test the external behavior of your package, not its internal implementation details.". So you are right that the default behavior is not mandatory and can be overridden.

marquiz (approved):
LGTM

PR #153, opened 2023-08-10
log the unresolved CDI devices only

tariq1890:
In `InjectDevices()` we are currently error-logging the whole list of devices instead of the device handles which weren't resolved

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -22,7 +22,7 @@ jobs:
       with:
         go-version: ${{ env.GO_VERSION }}
     - name: Install golint
-      run: go get -u golang.org/x/lint/golint
+      run: go install golang.org/x/lint/golint@latest
     - name: Lint
       run: make lint
     - name: Fmt
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -49,7 +49,7 @@ type Cache struct {
 }
 
 // WithAutoRefresh returns an option to control automatic Cache refresh.
-// By default auto-refresh is enabled, the list of Spec directories are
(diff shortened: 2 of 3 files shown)


tariq1890:
And some bonus changes to get the CI to play nice with me

elezar (approved):
Thanks @tariq1890.

One note. Is there a test that could be updated for this?

Not a blocker though.

.gitignore
```diff
@@ -1,3 +1,4 @@
 *.swp
 schema/validate
 /bin
+vendor/
```
klihub, line 4: nitpick: not a blocker necessarily, but this looks like an unrelated (and perhaps also unnecessary) change.

klihub (approved):
LGTM

.gitignore
elezar: Good point. @tariq1890 do you mind rolling back this change and the one in the sanity workflow?

Feel free to create a separate PR for these though.

tariq1890: Sure I can remove this change, but I think we need the change in the sanity workflow to get the CI checks to pass

.github/workflows/sanity.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       with:
         go-version: ${{ env.GO_VERSION }}
     - name: Install golint
-      run: go get -u golang.org/x/lint/golint
+      run: go install golang.org/x/lint/golint@latest
```
tariq1890, line 25: @elezar I was not able to get the CI checks to pass without this. I believe this is needed as a `go get` alters the go mod state

tariq1890:
@klihub would you be able to review this?

klihub (approved):
LGTM.

PR #158, opened 2023-09-01
pkg/cdi: remove github.com/opencontainers/runc/libcontainer dependency

thaJeztah:
- relates to https://github.com/cncf-tags/container-device-interface/pull/43
- relates to https://github.com/containerd/containerd/pull/9045

commit ee12044b3895afdff3841fee894e3edf01d7fc26 brought in this dependency on runc's libcontainer/devices package.

The runc maintainers consider "libcontainer" to be unstable, and not to be used by external consumers. The package (due to it being part of a large module) also brings in many transitive dependencies.

Given that the DeviceFromPath is only a small wrapper for some golang.org/x/sys calls, and the main purpose of the utility (giving information about "cgroup_permissions (which cannot be easily queried)") was not used, we can replace it with a more customised implemenation that's targetted at the information that's needed here.


thaJeztah:
@elezar PTAL 🤗

zvonkok:
Kicked off the tests; looks sane to me, and removing dependencies is always great. 

@thaJeztah Please check the failing tests, the PR looks incomplete regarding the go packages.

thaJeztah:
oh! hm.. interesting; let me check

thaJeztah:
Ah! only `go.sum`; looks like I forgot to stage that one before I push 😓 should be fixed now 🤞 can you kick CI again?

elezar:
I'm afk for a bit, but will look on Wednesday when I'm back at my desk. The failures in binary builds are probably due to the modules for the cmds needing updating. There are some vendor make targets that should do this.

zvonkok:
@elezar I will make sure the tests work and do a first pass, you can add your review when you're back.

zvonkok:
@thaJeztah  `cannot use stat.Mode (type uint32) as type uint16 in map index`

zvonkok:
Can you run the tests on your machine first?

thaJeztah:
Ah, dang; I know what's happening; the type differs between platforms (macOS uses different type than other Unix'es); will fix in a bit.

thaJeztah:
ok; changed it back to a switch, and ran tests in a container; I think it should be happy now

thaJeztah:
Hm... DCO check won't kick in for some reason 🤔

pkg/cdi/container-edits_unix.go
```diff
@@ -20,11 +20,42 @@
...
+	"errors"
 	"fmt"
 
-	runc "github.com/opencontainers/runc/libcontainer/devices"
+	"golang.org/x/sys/unix"
 )
 
+const (
+	blockDevice = "b"
+	charDevice  = "c" // or 'u'
```
elezar, line 31: nit: Should strictly be to match the `string` type:
```suggestion
	charDevice  = "c" // or "u"
```

elezar (approved):
@thaJeztah thanks for the cleanup. Removing the `runc` dependency here is great!

I think the DCO failure may be due to the migration of repos. Let me investigate.

klihub (approved):
LGTM.

thaJeztah:
> I think the DCO failure may be due to the migration of repos. Let me investigate.

If part of that was to mark the DCO check as "required", then I think you may be running into a bug in the DCO check.

What I ran into with that utility, is that it doesn't like marking it as a _required_ check if there are existing pull requests in the repository; in that case it will _never_ start.

A workaround for that is to make it temporarily "optional", and mark it "required" again when there are no open pull request (but, erm.. yeah, that's a bit of a pain on active repositories)

pkg/cdi/container-edits_unix.go
elezar: Let's update it for consistency if you don't mind.

elezar:
> > I think the DCO failure may be due to the migration of repos. Let me investigate.
> 
> If part of that was to mark the DCO check as "required", then I think you may be running into a bug in the DCO check.
> 
> What I ran into with that utility, is that it doesn't like marking it as a _required_ check if there are existing pull requests in the repository; in that case it will _never_ start.
> 
> A workaround for that is to make it temporarily "optional", and mark it "required" again when there are no open pull request (but, erm.. yeah, that's a bit of a pain on active repositories)

I unfortunately don't have permissions to add / edit the DCO check in this repository, so may have to bypass this for now. I have confirmed that your commit is signed-off as required.

cc @RobertKielty @krook

thaJeztah:
Let me do a quick push to address your comment https://github.com/cncf-tags/container-device-interface/pull/158#discussion_r1317189278

thaJeztah:
rebased, and addressed  👍 (needs "approval" again to run CI - sorry for that!)

elezar (approved):
LGTM. Thanks!

elezar:
merging bypassing the DCO check. Confirmed that commit was signed-off.

thaJeztah:
Thanks! Do you know if there were plans for a new tag? (I'll try to keep an eye, but if there's a new tag, I can update the version in containerd 😄)

krook:
> > > I think the DCO failure may be due to the migration of repos. Let me investigate.
> > 
> > 
> > If part of that was to mark the DCO check as "required", then I think you may be running into a bug in the DCO check.
> > What I ran into with that utility, is that it doesn't like marking it as a _required_ check if there are existing pull requests in the repository; in that case it will _never_ start.
> > A workaround for that is to make it temporarily "optional", and mark it "required" again when there are no open pull request (but, erm.. yeah, that's a bit of a pain on active repositories)
> 
> I unfortunately don't have permissions to add / edit the DCO check in this repository, so may have to bypass this for now. I have confirmed that your commit is signed-off as required.
> 
> cc @RobertKielty @krook

I've added the DCO bot at the org level for all repos. Please let me know if that fixes the problem going forward.

elezar:
> Thanks! Do you know if there were plans for a new tag? (I'll try to keep an eye, but if there's a new tag, I can update the version in containerd 😄)

Thanks @thaJeztah. I will check with the other members of the WG to tag and release the update.

PR #163, opened 2023-09-26
cmd/validate: print the validation errors

marquiz:
Print out any errors encountered during validation, as suggested by the command help.

### DIFF
--- a/cmd/cdi/cmd/validate.go
+++ b/cmd/cdi/cmd/validate.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -40,6 +41,13 @@ were reported by the registry.`,
 			return
 		}
 
+		fmt.Printf("CDI Registry has errors:\n")
+		for path, specErrors := range cdiErrors {
+			fmt.Printf("Spec file %s:\n", path)
+			for idx, err := range specErrors {
+				fmt.Printf("  %2d: %v\n", idx, strings.TrimSpace(err.Error()))
+			}
+		}
 		os.Exit(1)
 	},
 }


cmd/cdi/cmd/validate.go
```diff
@@ -40,6 +41,13 @@ were reported by the registry.`,
 			return
 		}
 
+		fmt.Printf("CDI Registry has errors:\n")
+		for path, specErrors := range cdiErrors {
+			fmt.Printf("Spec file %s:\n", path)
+			for idx, err := range specErrors {
+				fmt.Printf("  %d: %v\n", idx, strings.TrimRight(err.Error(), "\n"))
```
elezar, line 48: What about just:
```suggestion
				fmt.Printf("  %d: %v\n", idx, strings.Trim(err.Error()))
```
We should be ok to just ignore all leading and trailing whitespace in error messages.

elezar (commented):
One minor comment. Not a blocker though.

cmd/cdi/cmd/validate.go
marquiz: Makes sense. Changed it to `strings.TrimSpace()`

cmd/cdi/cmd/validate.go
```diff
@@ -40,6 +41,13 @@ were reported by the registry.`,
 			return
 		}
 
+		fmt.Printf("CDI Registry has errors:\n")
+		for path, specErrors := range cdiErrors {
+			fmt.Printf("Spec file %s:\n", path)
+			for idx, err := range specErrors {
+				fmt.Printf("  %d: %v\n", idx, strings.TrimSpace(err.Error()))
```
elezar, line 48: Thanks for the update. As a matter of interest, do we want to use tabs instead of spaces here so that the columns are aligned? This really isn't a blocker though, and I'm happy to approve as is.

elezar (commented):
minor nit but not a blocker.

elezar (approved):

cmd/cdi/cmd/validate.go
marquiz: Instead of using tabs I now changed this to `%3d` so it will reserve 3 chars for the idx. Will get unaligned if there are more than 999 errors

marquiz: Actually, that would be 999 errors per spec. I changed it to `%2d` so it will get unaligned after 99 errors per spec

Issue #162, opened 2023-09-15
Add support Edit Container Mount for kata/runtime-rs using direct volume in K8S/CSI scenario


Apokleos:
In the Kubernetes/CSI scenario, users need to use CSI to provide storage volumes to regular containers. After CSI completes the relevant operations, kubelet provides a HostPath and ContainerPath, but does not include a Mount Type. In other words, the Container Mount information that Containerd can retrieve through CRI does not include a type. When converting to the OCI Spec, the Mount type is set to "bind" by default.

Unfortunately,this solution does not work properly in the Kata Containers 3.0/runtime-rs with Direct Volume scenario.
When using a direct volume, Kata Pod needs to determine the type of DirectVolume based on Mount.Type after obtaining mount information. Then, it needs to choose the appropriate way to insert the backend device into the Guest, such as directvol, vfiovol, or spdkvol. Finally, the mount operation is completed in the Guest.  

Kata3.0/runtime-rs using direct volume with containerd ctr, more details can be seen [how-to-run-kata-containers-with-kinds-of-Block-Volumes](https://github.com/Apokleos/kata-containers/blob/multi-virtiofs/docs/how-to/how-to-run-kata-containers-with-kinds-of-Block-Volumes.md)
An example as below:
```
$ cat ./kubelet/kata-direct-vol-002/directvol002/mountInfo.json
{
  "device": "/tmp/stor/rawdisk01.20g", 
  "volume_type": "directvol", 
  "fs_type": "ext4", 
  "metadata":"{}", 
  "options": []
}
$ sudo ctr run -t --rm --runtime io.containerd.kata.v2 --mount type=directvol,src=/kubelet/kata-direct-vol-002/directvol002,dst=/disk002,options=rbind:rw "$image" kata-direct-vol-xx05302045 /bin/bash
```

However, there are certain differences between ordinary containers and Kata Containers when using CSI to access storage. The Mount Type in the OCI Spec provided by K8S/Containerd is by default the bind type. Ordinary containers will use the bind type by default, but Kata/DirectVolume requires a specific type, such as "directvol" (or other types like vfiovol or spdkvol). Therefore, in the K8S/CSI scenario, Kata Containers wants the Mount in its OCI Spec to contain a specific volume type instead of the bind type when creating a container that uses a Direct Volume. In other words, it is necessary to change the default bind type to the specific volume type at the right time, and then pass it to the Kata runtime-rs.

[...]

PR #169, opened 2023-10-09
Add support Edit ContainerMount for kata/direct volume

Apokleos:
Add support Edit Container Mount for kata/runtime-rs using direct volume in K8S/CSI scenario.

Fixes: #162

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -232,7 +232,10 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	for _, device := range devices {
 		d := c.devices[device]
 		if d == nil {
-			unresolved = append(unresolved, device)
+			// In kata direct-volume case, no need to unresolve it.
+			if !strings.HasPrefix(device, "kata.direct.volume") {
+				unresolved = append(unresolved, device)
+			}
 			continue
 		}
 		if _, ok := specs[d.GetSpec()]; !ok {
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -184,15 +184,13 @@ func ValidateDeviceName(name string) error {
 	for _, c := range string(name[1 : len(name)-1]) {
 		switch {
 		case IsAlphaNumeric(c):
-		case c == '_' || c == '-' || c == '.' || c == ':':
+		case c == '_' || c == '-' || c == '.' || c == ':' || c == '=':
 		default:
 			return fmt.Errorf("invalid character '%c' in device name %q",
 				c, name)
 		}
 	}
-	if !IsAlphaNumeric(rune(name[len(name)-1])) {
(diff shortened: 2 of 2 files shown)


pkg/cdi/cache.go
```diff
@@ -254,6 +259,19 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	return nil, nil
 }
 
+func (c *Cache) UpdateMounts(ociSpec *oci.Spec, mounts ...string) ([]string, error) {
```
elezar, line 262: This function is incorrectly named. It does not only update the mounts. It performs the same operations as `InjectDevices` with the exception that unresolved *devices* are output instead of returning an error.

In theory, I would not be against being able to control whether allowing unresolved devices when performing injection, but this is not the way I would propose to do it. Exposing an option at construction time and allowing a caller to retrieve the list of unresolved devices could be a possible solution.

pkg/cdi/cache.go
```diff
@@ -254,6 +259,19 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+	unresolved, edits, _ := c.filterDevices(ociSpec, mounts)
+	if unresolved != nil {
+		fmt.Printf("unresolvable CDI mount %s", strings.Join(mounts, ", "))
```
elezar, line 265: We should not just print to STDOUT here.

pkg/parser/parser.go
```diff
@@ -184,7 +184,7 @@ func ValidateDeviceName(name string) error {
 	for _, c := range string(name[1 : len(name)-1]) {
 		switch {
 		case IsAlphaNumeric(c):
-		case c == '_' || c == '-' || c == '.' || c == ':':
+		case c == '_' || c == '-' || c == '.' || c == ':' || c == '+' || c == '=':
```
elezar, line 187: Are we concerned with the fact that allowing an `=` could confict with the `vendor/class=name` pattern that we have for a fully-qualified CDI device name?

In addition, this was promted by the fact that we want to abuse the device name to pass base64-encoded data to something. These characters don't represent the set of valid base64 characters which includes `/`.

Note that this would also require a spec version bump.

elezar (changes requested):
@Apokleos thanks for the PR. 

As already pointed out in the linked issue, misusing the device name to pass data is not something that we can support in the specification. 

Let us continue the discussion there before we make a decision on this.

pkg/cdi/cache.go
Apokleos: Thx @elezar I have updated my code. Please review it again

pkg/cdi/cache.go
Apokleos: `UpdateMounts` removed

marquiz:
Rebased

PR #170, opened 2023-10-26
make specs-go a module

### LINKED ISSUE #97: reconsider dependencies
Using this repository is not possible in Kubernetes because even just having problematic dependencies in the dependency graph is flagged as a problem, even if no code from those dependencies is needed in Kubernetes:
- https://github.com/kubernetes/kubernetes/pull/111023#issuecomment-1309058680
- https://github.com/kubernetes/kubernetes/pull/111023#issuecomment-1309273728

The solution I came up with for now consists of copying some code from this repo into Kubernetes:
- test/e2e/dra/test-driver/app/cdi.go (simplified struct defs)
- pkg/kubelet/cm/dra/cdi.go (helper functions for working with CDI strings).

Long term it would be good to have that code in a Go module with no problematic dependencies.
This can be done by splitting this repo or with multiple Go modules in the existing one.

bart0sh:
This should allow to import CDI spec structures without adding a lot of CDI runtime deps.

It should partly fix #97  and allow Kubernetes to refer this module directly. There are at least 2 places in the k/k codebase that would benefit from this change:
- this can be replaced by the specs-go import: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dra/test-driver/app/cdi.go
- this can import specs-go and use JSON marshalling instead of quite ugly approach it uses to write CDI files: https://github.com/kubernetes/kubernetes/blob/master/test/images/sample-device-plugin/sampledeviceplugin.go#L237

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -13,7 +13,6 @@ github.com/hashicorp/go-multierror v1.1.1/go.mod h1:iw975J/qwKPdAO1clOe2L8331t/9
 github.com/mndrix/tap-go v0.0.0-20171203230836-629fa407e90b/go.mod h1:pzzDgJWZ34fGzaAZGFW22KVZDfyrYW+QABMrWnJBnSs=
 github.com/mrunalp/fileutils v0.5.0/go.mod h1:M1WthSahJixYnrXQl/DFQuteStB1weuxD2QJNHXfbSQ=
 github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
-github.com/opencontainers/runtime-spec v1.0.3-0.20220909204839-494a5a6aca78/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
 github.com/opencontainers/runtime-spec v1.1.0 h1:HHUyrt9mwHUjtasSbXSMvs4cyFxh+Bll4AjJ9odEGpg=
 github.com/opencontainers/runtime-spec v1.1.0/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
 github.com/opencontainers/runtime-tools v0.9.1-0.20221107090550-2e043c6bd626 h1:DmNGcqH3WDbV5k8OJ+esPWbqUOX5rMLR2PMvziDMJi0=
--- a/go.sum
+++ b/go.sum
@@ -13,6 +13,7 @@ github.com/hashicorp/go-multierror v1.1.1/go.mod h1:iw975J/qwKPdAO1clOe2L8331t/9
 github.com/mndrix/tap-go v0.0.0-20171203230836-629fa407e90b/go.mod h1:pzzDgJWZ34fGzaAZGFW22KVZDfyrYW+QABMrWnJBnSs=
 github.com/mrunalp/fileutils v0.5.0/go.mod h1:M1WthSahJixYnrXQl/DFQuteStB1weuxD2QJNHXfbSQ=
 github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
+github.com/opencontainers/runtime-spec v1.0.3-0.20220909204839-494a5a6aca78/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
 github.com/opencontainers/runtime-spec v1.1.0 h1:HHUyrt9mwHUjtasSbXSMvs4cyFxh+Bll4AjJ9odEGpg=
 github.com/opencontainers/runtime-spec v1.1.0/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
 github.com/opencontainers/runtime-tools v0.9.1-0.20221107090550-2e043c6bd626 h1:DmNGcqH3WDbV5k8OJ+esPWbqUOX5rMLR2PMvziDMJi0=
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -13,6 +13,7 @@ github.com/hashicorp/go-multierror v1.1.1/go.mod h1:iw975J/qwKPdAO1clOe2L8331t/9
 github.com/mndrix/tap-go v0.0.0-20171203230836-629fa407e90b/go.mod h1:pzzDgJWZ34fGzaAZGFW22KVZDfyrYW+QABMrWnJBnSs=
 github.com/mrunalp/fileutils v0.5.0/go.mod h1:M1WthSahJixYnrXQl/DFQuteStB1weuxD2QJNHXfbSQ=
 github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
+github.com/opencontainers/runtime-spec v1.0.3-0.20220909204839-494a5a6aca78/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
```
bart0sh, line 16: NOTE: this is the version used in k/k: https://github.com/kubernetes/kubernetes/blob/v1.29.0-alpha.2/go.mod#L206
I guess a lot of deps were removed from cmd/cdi/go.sum because of this change, which looks rather good.

elezar:
One question: Does this mean we need to update the `mod-tidy` and `mod-verify` targets in the `Makefile` to include the `specs-go` folder too?

klihub:
So I guess after merging this we should start tagging also the specs-go golang sub-module ?

klihub:
> One question: Does this mean we need to update the `mod-tidy` and `mod-verify` targets in the `Makefile` to include the `specs-go` folder too?

Yes, I think so. Maybe we should replace the current, manually spelt out tidying and verifying rules with something more simple and automatic like, for instance:

```diff
diff --git a/Makefile b/Makefile
index 6fc1836..9f70ee7 100644
--- a/Makefile
+++ b/Makefile
@@ -64,30 +64,18 @@ $(BINARIES): bin/%:
 #
 # go module tidy and verify targets
 #
-.PHONY: mod-tidy $(CMD_MOD_TIDY_TARGETS) mod-tidy-root
-.PHONY: mod-verify $(CMD_MOD_VERIFY_TARGETS) mod-verify-root
-
-CMD_MOD_TIDY_TARGETS := mod-tidy-cdi mod-tidy-validate
-CMD_MOD_VERIFY_TARGETS := mod-verify-cdi mod-verify-validate
-
-mod-tidy-root:
-       $(Q)echo "Running $@..."; \
-       $(GO_CMD) mod tidy
-
-$(CMD_MOD_TIDY_TARGETS): mod-tidy-%: mod-tidy-root
-       $(Q)echo "Running $@... in $(abspath ./cmd/$(*))"; \
-       (cd $(abspath ./cmd/$(*)) && $(GO_CMD) mod tidy)
-
-mod-verify-root: mod-tidy-root
-       $(Q)echo "Running $@..."; \
-       $(GO_CMD) mod verify
-
-$(CMD_MOD_VERIFY_TARGETS): mod-verify-%: mod-tidy-% mod-verify-root
-       $(Q)echo "Running $@... in $(abspath ./cmd/$(*))"; \
-       (cd $(abspath ./cmd/$(*)) && pwd && $(GO_CMD) mod verify)
-
-mod-verify: $(CMD_MOD_VERIFY_TARGETS)
-mod-tidy: $(CMD_MOD_TIDY_TARGETS)
+.PHONY: mod-tidy
+.PHONY: mod-verify
+
+mod-tidy:
+       $(Q)for mod in $$(find . -name go.mod); do \
+           (echo "Tidying $$mod..."; cd $$(dirname $$mod) && go mod tidy) || exit 1; \
+       done
+
+mod-verify:
+       $(Q)for mod in $$(find . -name go.mod); do \
+           (echo "Veriying $$mod..."; cd $$(dirname $$mod) && go mod verify) || exit 1; \
+       done
 
 #
 # cleanup targets
```

klihub (commented):
@bart0sh I'm fine with turning this into a sub-module of its own.

klihub:
> And in addition to doing so, I would also add github workflow job to reject any PR where a `make mod-tidy` would introduce some changes...
> 
> @elezar @bart0sh If you guys agree, I can file a PR to these effects...

@elezar @bart0sh #171

bart0sh:
@klihub 
> So I guess after merging this we should start tagging also the specs-go golang sub-module ?

That would make sense from my point of view.

bart0sh:
@klihub @elezar considering #171 lgtm/approve anyone?

klihub (approved):
LGTM

elezar:
> @klihub
> 
> > So I guess after merging this we should start tagging also the specs-go golang sub-module ?
> 
> That would make sense from my point of view.

What does "tagging the sub-module" mean in this context. We only have a single tag associated with this repository.

pohly:
> What does "tagging the sub-module" mean in this context.

https://go.dev/doc/modules/managing-source#multiple-module-source

elezar:
> > What does "tagging the sub-module" mean in this context.
> 
> https://go.dev/doc/modules/managing-source#multiple-module-source

OK. So if we create a git tag: `specs-go/v0.6.0` this will effectively version the `specs-go` submodule and allow consumers to import this? Is importing from the root still supported, or would this change require that our consumers need to be updated?

specs-go/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/container-orchestrated-devices/container-device-interface/specs-go/v0.0.0
```
pohly, line 1: This version at the end is surprising. Is it needed?

klihub: No, I think it is a mistake to have it there.

klihub (changes requested):
@bart0sh Could you fix the go.mod thinko @pohly spotted ?

specs-go/go.mod
bart0sh: fixed. thank you for pointing out.

bart0sh:
> @bart0sh You mean this one or #171 itself ?

this one. I don't mind merging #171 before this or after.

klihub (approved):

elezar:
/lgtm

@bart0sh for your info. I rebased on `main` and fixed the module definition for `cmd/validate`.

PR #171, opened 2023-10-26
.github, Makefile: sanity workflow fixes, simplify mod-* targets, bump golang to 1.19.x

klihub:
This PR
- simplifies go.{mod,sum} tidying and fixes existing deviations from tidy go mods
- fixes the existing gofmt check in the sanity workflow and existing gofmt violations in the code
- adds go mod tidiness verification to the sanity workflow
- bumps go version to 1.19.x everywhere
- updates go deps


.github/workflows/sanity.yml
```diff
@@ -29,6 +29,9 @@ jobs:
       run: make fmt
     - name: Vet
       run: make vet
+    - name: Verify go modules are properly up-to-date
+      run: |
+        make mod-tidy mod-verify && git diff --exit-code
```
bart0sh, line 35: Can you explain why do we need `git diff --exit-code` here? As far as I understood `make mod-tidy mod-verify` returns proper exit code, no?

klihub: mod-tidy / go mod tidy exits with a 0 status on success including cases where it needed to update `go.{mod,sum}` but succeeded in doing so. In contrast, here we want to make sure that `go.{mod,sum}` was already up-to-date, or IOW, there was no need to touch and modify them. `git diff --exit-code` is there to make sure that any (successful) changes made to these files are considered an error and not a success.

.github/workflows/sanity.yml
```diff
@@ -26,9 +26,13 @@ jobs:
...
       run: make lint
     - name: Fmt
-      run: make fmt
+      run: |
+        make fmt && git diff --exit-code
     - name: Vet
       run: make vet
+    - name: Verify go modules are properly up-to-date
+      run: |
+        make mod-tidy mod-verify && git diff --exit-code
```
elezar, line 35: One question: Is this guaranteed to run `mod-tidy` and then `mod-verify`?

elezar (commented):
Thanks for the simplification @klihub.

One minor question on the sanity workflow, but otherwise this looks good.

.github/workflows/sanity.yml
bart0sh: btw, would it make sense to merge `mod-tidy` and `mod-verify` into one target? They seem to always used together.

.github/workflows/sanity.yml
bart0sh: Thank you. lgtm then.

.github/workflows/sanity.yml
elezar: `mod-tidy` can also be used during development before checking things in, so it may make sense to keep them separate.

bart0sh: > mod-tidy can also be used during development before checking things in

Considering that there is a job that runs both targets, what's the point to run only one before checking things in?

klihub: > One question: Is this guaranteed to run `mod-tidy` and then `mod-verify`?

You mean if both are run, is it guaranteed that they are run in that order ? Yes, unless there is a direct or indirect dependency between the two targets, in which case they would be always run in dependency order.

klihub: And for us, those are both phony targets without any prereqs/dependencies.

klihub: Actually, I have to say that now that I looked at `go mod help verify`, I'm not sure if it makes any sense to run `go mod verify` in a workflow job if/where everything start afresh from scratch...

elezar: Yes, I think `make mod-tidy` is sufficient in this case.

klihub: Okay, I'll update the PR...

klihub: Updated.

elezar (approved):
/lgtm

PR #173, opened 2023-10-30
Add version to spec-go dependency

elezar:
This change adds a v0.6.0 version to the internal specs-go dependency. This prevents a transitive dependency on the v0.0.0-00010101000000-000000000000 pseudo version when importing the package into clients.


klihub (approved):
LGTM.

PR #182, opened 2024-01-10
test: add missing error checks in cache tests

marquiz:


pkg/cdi/cache_test.go
```diff
@@ -554,6 +558,7 @@ devices:
 							opts = append(opts, WithAutoRefresh(false))
 						}
 						cache, err = NewCache(opts...)
+						require.Nil(t, err)
```
elezar, line 561: Nit: replace with:
```suggestion
						require.NoError(t, err)
```

marquiz: Thanks for pointing this out. Changed as suggested

elezar (approved):

PR #183, opened 2024-01-10
pkg/cdi: add missing error checks in cache

marquiz:
Mostly retain the existing behavior of ignoring errors but using `_` to make golangci-lint happy.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -69,7 +69,9 @@ func NewCache(options ...Option) (*Cache, error) {
 		watch:       &watch{},
 	}
 
-	WithSpecDirs(DefaultSpecDirs...)(c)
+	if err := WithSpecDirs(DefaultSpecDirs...)(c); err != nil {
+		return nil, err
+	}
 	c.Lock()
 	defer c.Unlock()
 
@@ -107,9 +109,7 @@ func (c *Cache) configure(options ...Option) error {
 		c.watch.setup(c.specDirs, c.dirErrors)
 		c.watch.start(&c.Mutex, c.refresh, c.dirErrors)
 	}
-	c.refresh()
-
-	return nil
+	return c.refresh()
 }
 
 // Refresh rescans the CDI Spec directories and refreshes the Cache.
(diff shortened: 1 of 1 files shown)


marquiz:
I was not completely confident I'm doing the right thing here. Thus draft :) As a general note I think the very common pattern here of omitting error checking is bad design. Either check it or then don't even return an error in the first place.

elezar:
Syntactically, I think this is fine. I'll defer to @klihub for when we should actually return the errors when refreshing the cache.

marquiz:
> Syntactically, I think this is fine. I'll defer to @klihub

Yeah, no syntax error 😇 ping @klihub

pkg/cdi/cache.go
```diff
@@ -69,7 +69,9 @@ func NewCache(options ...Option) (*Cache, error) {
 		watch:       &watch{},
 	}
 
-	WithSpecDirs(DefaultSpecDirs...)(c)
+	if err := WithSpecDirs(DefaultSpecDirs...)(c); err != nil {
```
klihub, line 72: WithSpecDirs can't fail, so this is fine.

pkg/cdi/cache.go
```diff
@@ -107,9 +109,7 @@ func (c *Cache) configure(options ...Option) error {
 		c.watch.setup(c.specDirs, c.dirErrors)
 		c.watch.start(&c.Mutex, c.refresh, c.dirErrors)
 	}
-	c.refresh()
-
-	return nil
+	return c.refresh()
```
klihub, line 112: TBH, off the top of my head I'm not sure if this could cause problems.

Looking at the code, maybe it would be better/simpler/cleaner to change the API so that [Rr]efresh() never returns an error. Since errors encountered during the last refresh can anyway be queried using GetErrors(), if a caller is interested in errors it can query them itself.

pkg/cdi/cache.go
marquiz: Hmm, ok, this lead me to submit #187 (replaces this particular change)

elezar:
@marquiz this can be closed now, correct?

marquiz:
We agreed to not break the Cache API (return an error from `Configure()` even though it's always `nil`), so I'll close this one

PR #186, opened 2024-01-10
github: add golangci-lint in the sanity CI job

marquiz:


marquiz:
Should pass muster when #181, #182, #183, #184 and #185 have been merged

PR #187, opened 2024-01-11
chore: simplify cache Option type

marquiz:
Options never fail so why return an error.


elezar (approved):
Thanks for the cleanup @marquiz

PR #188, opened 2024-01-11
pkg/cdi: drop unused return value from NewCache function

marquiz:
Drop the unused (always nil) error returned by NewCache(). This eliminate some unreachable error paths in the code.

> **NOTE:** This change changes the exported API of "pkg/cdi" package as the prototype of NewCache() changes.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -62,7 +62,7 @@ func WithAutoRefresh(autoRefresh bool) Option {
 // NewCache creates a new CDI Cache. The cache is populated from a set
 // of CDI Spec directories. These can be specified using a WithSpecDirs
 // option. The default set of directories is exposed in DefaultSpecDirs.
-func NewCache(options ...Option) (*Cache, error) {
+func NewCache(options ...Option) *Cache {
 	c := &Cache{
 		autoRefresh: true,
 		watch:       &watch{},
@@ -72,7 +72,8 @@ func NewCache(options ...Option) (*Cache, error) {
 	c.Lock()
 	defer c.Unlock()
 
-	return c, c.configure(options...)
+	c.configure(options...)
+	return c
 }
 
 // Configure applies options to the Cache. Updates and refreshes the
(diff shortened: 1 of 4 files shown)


marquiz:
#187 was merged

pkg/cdi/cache.go
```diff
@@ -62,7 +62,7 @@ func WithAutoRefresh(autoRefresh bool) Option {
 // NewCache creates a new CDI Cache. The cache is populated from a set
 // of CDI Spec directories. These can be specified using a WithSpecDirs
 // option. The default set of directories is exposed in DefaultSpecDirs.
-func NewCache(options ...Option) (*Cache, error) {
+func NewCache(options ...Option) *Cache {
```
elezar, line 65: Just one note: This will require that `moby` be updated to use the new API: https://github.com/moby/moby/blob/92cc0d62570098dd6b4cfadeb3982e0157578735/daemon/cdi.go#L63

elezar (commented):
One minor comment on a breaking API change. 

I'm happy to update `moby` once we publish this though. It would be good to call this out in the issue description though.

pkg/cdi/cache.go
marquiz: Ach, ok, I thought registry is the "real" external API. We don't need to break this API if so desired 😅

klihub: > > Ach, ok, I thought registry is the "real" external API. We don't need to break this API if so desired 😅

This breaks the API now and also, if in the future we need to introduce a new option that can genuinely fail this change would force us to break the API then again. That said, since we haven't been even talking about any new options, if you guys really want to make this change, I guess it should be fine.

> The moby devs had an issue with the singleton pattern for the registry and requested that we instantiate the cache directly. I usually bump the `moby` dependency anyway, so let's just make sure that this makes it into the releaes notes. (Maybe update the commit message and PR description to indicate the breaking change).

I have to admit that I also hate our current superficial registry API. Not the singleton pattern as such, but I'd rather go back to the original proposed golang stdlib/flag-like API, where all cache-level functions were available as package level functions and they were operating on the implicitly available (singleton) default cache instance.

Anyway, @elezar @marquiz let me know if you'd like to proceed with this and I can LGTM it then.

marquiz:
Thanks for the reviews. I got stuck pondering that if we change the cache API should we change registry API correspondingly, or not?

pkg/cdi/cache.go
marquiz: > (Maybe update the commit message and PR description to indicate the breaking change

K, I now updated the commit message (title and body) and the PR title, accordingly. PTAL @elezar

klihub (approved):
LGTM

elezar (approved):
Thanks for the discussion here @marquiz and @klihub.

LGTM

PR #189, opened 2024-01-18
pkg/cdi: drop unused return value from Configure methods

marquiz:
**NOTE:** This change changes the exported API of "pkg/cdi" package as the
prototype of Configure() method of Cache and Registry types changes.

A follow-up on top of #188

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -78,17 +78,13 @@ func NewCache(options ...Option) *Cache {
 
 // Configure applies options to the Cache. Updates and refreshes the
 // Cache if options have changed.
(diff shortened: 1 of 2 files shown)


elezar:
We discussed this change in the COD working group meeting and would suggest that we *do not* merge it. See https://docs.google.com/document/d/1gUgAMEThkRt4RJ7pA7ZbPPmIOX2Vb7fwH025MjfcTYU/edit#bookmark=id.khpjizwm6m54

We already have clients that use this function (Containerd, Cri-o) and this would require explicit code changes on their end. Furthermore, it is reasonable to assume that we could require `Option`s that return errors at some point.

PR #190, opened 2024-01-23
Revert NewCache function signature to return an error

elezar:
This change reverts the NewCache function signature to also return an error. This was removed in #188 but does break clients such as moby using the NewCache API.

Instead, a private function is added that returns no error and this is used internally and in tests.


klihub (approved):
LGTM.

marquiz (approved):
LGTM

PR #195, opened 2024-03-01
pkg/cdi: deprecate Registry and related interfaces.

klihub:
Mark Registry and its interfaces deprecated.


elezar (approved):
LGTM.

Do we want @marquiz to have a look too?

PR #196, opened 2024-03-01
pkg,schema,internal: eliminate multierror.

klihub:
Migrate from our internal multierror implementation to using error slices and stock errors.Join().


elezar (commented):
I think we need to bump our go version to at least 1.19.

klihub:
> I think we need to bump our go version to at least 1.19.

Yeah, I figured too.

elezar:
Are our clients using go >= 1.21 to build their applications?

For completeness, I recently did something similar in one of our projects to enable `errors.Join` in go1.17: https://github.com/NVIDIA/nvidia-container-toolkit/pull/337

pkg/cdi/cache.go
```diff
@@ -133,11 +132,11 @@ func (c *Cache) Refresh() error {
...
 
 	// collect and return cached errors, much like refresh() does it
-	var result error
+	errs := []error{}
 	for _, errors := range c.errors {
-		result = multierror.Append(result, errors...)
+		errs = append(errs, errors...)
 	}
-	return result
+	return errors.Join(errs...)
```
elezar, line 139: var errs error
for _, errors := range c.errors {
    errs = errors.Join(errs, errors)
}
```

File: pkg/cdi/cache.go [hunk 2]
```diff
@@ -147,12 +146,12 @@ func (c *Cache) refresh() error {
 		devices    = map[string]*Device{}
 		conflicts  = map[string]struct{}{}
 		specErrors = map[string][]error{}
-		result     []error
+		errs       = []error{}
```
elezar, line 149:
```suggestion
		errs       error
```

pkg/cdi/cache.go
```diff
@@ -147,12 +146,12 @@ func (c *Cache) refresh() error {
...
 	)
 
 	// collect errors per spec file path and once globally
 	collectError := func(err error, paths ...string) {
-		result = append(result, err)
+		errs = append(errs, err)
```
elezar, line 154:
```suggestion
		errs = errors.Join(errs, err)
```

pkg/cdi/cache.go
```diff
@@ -205,7 +204,7 @@ func (c *Cache) refresh() error {
 	c.devices = devices
 	c.errors = specErrors
 
-	return multierror.New(result...)
+	return errors.Join(errs...)
```
elezar, line 207:
```suggestion
	return errs
```

schema/schema.go
```diff
@@ -333,11 +333,16 @@ func (e *Error) Error() string {
 		return ""
 	}
 
-	var multi error
+	errs := []error{}
 	for _, err := range e.Result.Errors() {
-		multi = multierror.Append(multi, fmt.Errorf("%v", err))
+		errs = append(errs, fmt.Errorf("%v", err))
```
elezar, line 338:
```suggestion
		errs = errors.Join(errs, err)
```
(this assumes that `.Errors()` returns `error` and not `string` -- otherwise the `fmt.Errorf` is still required.

schema/schema.go
```diff
@@ -333,11 +333,16 @@ func (e *Error) Error() string {
...
 	}
-	return multi.Error()
+
+	if err := errors.Join(errs...); err != nil {
+		return fmt.Sprintf("%v", err)
+	}
```
elezar, line 343:
```suggestion
	if errs != nil {
		return fmt.Sprintf("%v", errs)   
	}
```

internal/validation/k8s/objectmeta.go
```diff
@@ -20,28 +20,27 @@ limitations under the License.
...
 
 // ValidateAnnotations validates that a set of annotations are correctly defined.
 func ValidateAnnotations(annotations map[string]string, path string) error {
-	errors := multierror.New()
+	errs := []error{}
 	for k := range annotations {
 		// The rule is QualifiedName except that case doesn't matter, so convert to lowercase before checking.
 		for _, msg := range IsQualifiedName(strings.ToLower(k)) {
-			errors = multierror.Append(errors, fmt.Errorf("%v.%v is invalid: %v", path, k, msg))
+			errs = append(errs, fmt.Errorf("%v.%v is invalid: %v", path, k, msg))
```
elezar, line 37:
```suggestion
			errs = errors.Join(errs, fmt.Errorf("%v.%v is invalid: %v", path, k, msg))
```

internal/validation/k8s/objectmeta.go
```diff
@@ -20,28 +20,27 @@ limitations under the License.
...
 		}
 	}
 	if err := ValidateAnnotationsSize(annotations); err != nil {
-		errors = multierror.Append(errors, fmt.Errorf("%v is too long: %v", path, err))
+		errs = append(errs, fmt.Errorf("%v is too long: %v", path, err))
```
elezar, line 41:
```suggestion
		errs = errors.Join(errs, fmt.Errorf("%v is too long: %v", path, err))
```

internal/validation/k8s/objectmeta.go
```diff
@@ -20,28 +20,27 @@ limitations under the License.
...
 	}
-	return errors
+	return errors.Join(errs...)
```
elezar, line 43:
```suggestion
	return errs
```

elezar (commented):
The use of `errors.Join` can clean this up even further since it's never required to store a slice of errors.

internal/validation/k8s/objectmeta.go
klihub: That is not the same.

While I realize that from a pure compilability/errors API point of view it could be implemented like that, I deliberate chose not to do it.

We've converted a number of packages from multierror (both the original hashicorp one and one-off imitation like this) and in every conversion so far we've tried to consistently follow this same pattern: keep errors that originate at the same 'level' also at the same level in the joint error.

If you look at the errors.Join, you'll see that your suggested changes end up with a very different internal representation. Instead of a flat joinError with a single slice containing all the errors at the same logical level, you end up with deeply nested (sort of unbalanced tree) of joinErrors, with each one containing a single error as errs[0] and a similarly recursive jointError as errs[1].

pkg/cdi/cache.go
klihub: Yeah, actually I did this wrong as I flattened all errors to a single joinError. I updated it to return a joinError of one joinError per Spec errors.

schema/schema.go
klihub: No, unfortunately Result.Errors() returns []ResultError which is neither error nor []error. It implements String() but not Error(), so we need to turn it into a error using fmt.Errorf.

klihub:
> The use of `errors.Join` can clean this up even further since it's never required to store a slice of errors.

If you don't care about how the topology of the final set of stored errors look like, but (for some reason) I prefer to keep related errors in the same slice and only resulting errors (errors caused by other errors) nested the way it would be if we use the code you suggested. See my other related comment about the details and reasoning.

pkg/cdi/cache.go
klihub: Yeah, this wasn't producing a natural collection/topology of joint errors. Updated.

pkg/cdi/cache.go
klihub: Yeah, I did this wrong, as it resulted in a flat topology, whereas a joinError of one joinErrors per Spec file wrapping errors of that file only is more natural. Updated like that.

internal/validation/k8s/objectmeta.go
```diff
@@ -20,28 +20,27 @@ limitations under the License.
...
-	"tags.cncf.io/container-device-interface/internal/multierror"
 )
 
 // TotalAnnotationSizeLimitB defines the maximum size of all annotations in characters.
 const TotalAnnotationSizeLimitB int = 256 * (1 << 10) // 256 kB
 
 // ValidateAnnotations validates that a set of annotations are correctly defined.
 func ValidateAnnotations(annotations map[string]string, path string) error {
-	errors := multierror.New()
+	errs := []error{}
```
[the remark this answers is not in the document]
klihub: That does not result in the joinError/topology I was after: one (join)Error wrapping in a single slice of errors, all annotation validation errors for that path.

elezar:
> > The use of `errors.Join` can clean this up even further since it's never required to store a slice of errors.
> 
> If you don't care about how the topology of the final set of stored errors look like, but (for some reason) I prefer to keep errors 'at the same level' in the same slice (of joinErrors) and only 'cascading result of errors' (errors caused by other errors) nested the way it would be if we used the code you suggested. See my other related comment...

OK. That's a fair point. I was not considering this aspect.

elezar (commented):
I think I'm good with the changes now. Thanks for motivating your choices so clearly.

My final concern is the go version bump. Do we foresee issues with clients consuming this module?

cmd/cdi/go.mod
```diff
@@ -1,6 +1,8 @@
 module tags.cncf.io/container-device-interface/cmd/cdi
 
-go 1.19
+go 1.21
```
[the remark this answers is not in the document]
klihub: cri-o and containerd are already at 1.21 but podman looks to be at 1.20. So if 1.20 proves to be recent enough for this PR, then it can easily be a better choice. I'll update the PR and then we'll see.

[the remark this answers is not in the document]
klihub: Okay, everything looks fine with 1.20.

elezar (approved):
Thanks @klihub. I think I'm fine with the 1.20 bump here.

Once Podman bumps their version we could look at doing that here too.

PR #197, opened 2024-03-07
Allow top-level Configure to return an error

elezar:


marquiz (approved):
LGTM. Makes sense to me

PR #202, opened 2024-04-16
Decrease GitHub.com/opencontainers/runtime spec

### LINKED ISSUE #201: Decrease OCI runtime spec dependency version
The recent 0.7.1 release includes a bump of the `github.com/opencontainers/runtime-spec` dependency to `v1.2.0`. This has caused some issues on downstream dependencies and it may be better to decreast the required version to `v1.0.0` and also remove the dependabot update for these dependencies.

See:
* https://github.com/containerd/containerd/pull/10077#discussion_r1567091564
* https://github.com/cri-o/cri-o/pull/7971

elezar:
Since we are essenitally a library, we need to specify the MINIMUM version of the OCI runtime specification that we need. 

This change:
* Drops the required OCI runtime spec version in the CDI spec to v1.1.0 (where the `IntelRdt.EnableCMT` and `IntelRdt.EnableMBM` fields wer added).
* Removes the dependabot update for the OCI runtime spec.

closes #201

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,14 +1,5 @@
 version: 2
 updates:
-- package-ecosystem: gomod
-  directory: "/"
-  schedule:
-    interval: weekly
-  open-pull-requests-limit: 10
-  allow:
-  - dependency-name: "github.com/opencontainers/runtime-spec"
-    dependency-type: direct
-
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
--- a/cmd/cdi/go.mod
+++ b/cmd/cdi/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/fsnotify/fsnotify v1.5.1
-	github.com/opencontainers/runtime-spec v1.2.0
+	github.com/opencontainers/runtime-spec v1.1.0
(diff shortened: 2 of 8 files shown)


marquiz (commented):
Makes sense. Should we roll back the version in `cmd/cdi`, too?

elezar:
I'm not too concerned with the cmd dependencies since they're not imported into downstream dependencies. I suppose it does make sense to update them though.

elezar:
@marquiz I have updated the module at `cmd/cdi` too. 

@zvonkok if I recall correctly you may have been using this functionality for some prototyping. Does that depend on a newer spec version?

marquiz:
> I'm not too concerned with the cmd dependencies since they're not imported into downstream dependencies. I suppose it does make sense to update them though.

Yeah, I'm not that concerned either, just for consistency as I noticed that cmd/validate was updated, too

marquiz (approved):
Looks good, thanks @elezar

zvonkok (approved):
LGTM

klihub (approved):

elezar:
@klihub @marquiz just realized that we actually need to the v1.1.0 spec.

klihub (approved):

klihub:
Fair enough, LGTM.

Issue #206, opened 2024-05-14
Drive CDI specification to v1.0.0


elezar:
This issue serves as a top-level tracker for achieving a v1.0.0 CDI specification.

See the discussions here:
* [May 16 2023](https://docs.google.com/document/d/1gUgAMEThkRt4RJ7pA7ZbPPmIOX2Vb7fwH025MjfcTYU/edit#bookmark=id.dbq5h7af0wfc)
* [April 16 2024](https://docs.google.com/document/d/1gUgAMEThkRt4RJ7pA7ZbPPmIOX2Vb7fwH025MjfcTYU/edit#bookmark=id.w0eos1p8964q)

- [x] https://github.com/cncf-tags/container-device-interface/pull/214
- [x] #215 
- [x] #231
  - Currently we fail on unknown fields and then we check whether the version field in the spec file is supported by the  consumer. We also use golang null default values for all new fields so that these remain valid when loading older spec versions.
- [ ] #230
  - [ ] producer: Minimum spec version: https://github.com/NVIDIA/nvidia-container-toolkit/blob/f794d09df19b628232d872470772b50f34b45ae8/pkg/nvcdi/spec/set-minimum-version.go#L28
- [x] Remove deprecated APIs
  - [x] #218 
  - [x] #219
- [x] Remove dependency on runtime-spec from spec-go for producer API
  - [x] https://github.com/cncf-tags/container-device-interface/pull/208
- [x] Code cleanup
  - [x] #222
  - [x] #223 
  - [x] #227
- [x] #235
- [ ] [nice-to-have] Automated checks for parity between Go and Rust implementations
- [ ] [nice-to-have] Add reference implementation for hooks
  - [ ] #225

bart0sh:
@elezar @klihub  should we also remove all deprecated APIs in v1.0.0 ?

elezar:
> @elezar @klihub should we also remove all deprecated APIs in v1.0.0 ?

Yes, that sounds reasonable.

klihub:
> > @elezar @klihub should we also remove all deprecated APIs in v1.0.0 ?
> 
> Yes, that sounds reasonable. 

Yup, we can do that, too.

bart0sh:
[...]

PR #214, opened 2024-07-29
Move version APIs to the specs-go package

bart0sh:
This is a partial fix for #206


bart0sh:
@elezar @klihub PTAL

klihub (approved):
LGTM.

bart0sh:
@elezar can you review this PR?

specs-go/config.go
```diff
@@ -2,9 +2,6 @@ package specs
 
 import "os"
 
-// CurrentVersion is the current version of the Spec.
```
elezar, line 5: I think we should keep the version defined here.

specs-go/version.go
```diff
@@ -60,50 +58,58 @@ var validSpecVersions = requiredVersionMap{
...
-func MinimumRequiredVersion(spec *cdi.Spec) (string, error) {
+func MinimumRequiredVersion(spec *Spec) (string, error) {
 	minVersion := validSpecVersions.requiredVersion(spec)
 	return minVersion.String(), nil
 }
 
-// version represents a semantic version string
-type version string
+// Version represents a semantic version string
+type Version string
```
elezar, line 76: If we only expose `ValidateVersion` and `MinimumRequiredVersion` as the APIs here, we don't need to make this public.

elezar (changes requested):
Thanks @bart0sh.

Definitely a great start. I think we can simplify the API that we expose w.r.t the versions and limit it to two functions:
* `ValidateVersion(*Spec) error`
* `MinimumRequiredVersion(*Spec) (string, error)`

specs-go/version.go
bart0sh: Thank you for pointing out! done.

bart0sh: BTW, this can be also considered as an API break as this type was public.

pkg/cdi/spec.go
```diff
@@ -203,15 +203,15 @@ func (s *Spec) edits() *ContainerEdits {
 
 // Validate the Spec.
 func (s *Spec) validate() (map[string]*Device, error) {
-	if err := validateVersion(s.Version); err != nil {
+	if err := cdi.ValidateVersion(s.Version); err != nil {
 		return nil, err
 	}
 
-	minVersion, err := MinimumRequiredVersion(s.Spec)
```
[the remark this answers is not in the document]
bart0sh: done

specs-go/parser.go
```diff
@@ -0,0 +1,39 @@
...
+import "strings"
+
+// ParseQualifier splits a device qualifier into vendor and class.
+// The syntax for a device qualifier is
+//
+//	"<vendor>/<class>"
+//
+// If parsing fails, an empty vendor and the class set to the
+// verbatim input is returned.
+func ParseQualifier(kind string) (string, string) {
```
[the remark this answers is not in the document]
bart0sh: I'm not sure I follow. `Parser.ParseQualifier` already calls `cdi.ParseQualifier`. Can you elaborate?

pkg/cdi/spec.go
```diff
@@ -203,15 +203,15 @@ func (s *Spec) edits() *ContainerEdits {
...
+	minVersion, err := cdi.MinimumRequiredVersion(s.Spec)
 	if err != nil {
 		return nil, fmt.Errorf("could not determine minimum required version: %v", err)
 	}
-	if newVersion(minVersion).IsGreaterThan(newVersion(s.Version)) {
+	if cdi.NewVersion(minVersion).IsGreaterThan(cdi.NewVersion(s.Version)) {
 		return nil, fmt.Errorf("the spec version must be at least v%v", minVersion)
 	}
```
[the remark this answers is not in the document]
bart0sh: This would introduce a circular dependency from spec-go to cdi package as far as I understand.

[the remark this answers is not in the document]
bart0sh: Ah, got it. Should be done now.

pkg/cdi/spec_test.go
```diff
@@ -525,7 +525,7 @@ func specType(content []byte) string {
 }
 
 func TestCurrentVersionIsValid(t *testing.T) {
-	require.NoError(t, validateVersion(cdi.CurrentVersion))
+	require.NoError(t, cdi.ValidateVersion(&cdi.Spec{Version: cdi.CurrentVersion}))
 }
```
elezar, line 529: Should this now be moved to `specs-go/version_test.go`?

pkg/cdi/spec_test.go
```diff
@@ -525,7 +525,7 @@ func specType(content []byte) string {
...
 
 func TestRequiredVersion(t *testing.T) {
```
elezar, line 531: We could consider moving these to `specs-go/version_test.go`.

specs-go/config.go
elezar: I'm OK to move it to `version.go` too. Thanks for confirming the behaviour in the OCI runtime spec.

specs-go/parser.go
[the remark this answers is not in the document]
elezar: Sorry, I just saw that this was the case. I think I missed it in my initial pass.

specs-go/version.go
```diff
@@ -14,36 +14,34 @@
...
-
-	"tags.cncf.io/container-device-interface/pkg/parser"
-	cdi "tags.cncf.io/container-device-interface/specs-go"
 )
 
 const (
-	// CurrentVersion is the current version of the CDI Spec.
-	CurrentVersion = cdi.CurrentVersion
+	// CurrentVersion is the current version of the Spec.
+	CurrentVersion = "0.8.0"
```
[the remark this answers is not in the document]
elezar: OK. Let's leave it here then. Thanks.

specs-go/version.go
```diff
@@ -60,13 +58,28 @@ var validSpecVersions = requiredVersionMap{
 	v080: requiresV080,
 }
 
+// ValidateVersion checks whether the specified spec version is supported.
```
elezar, line 61:
```suggestion
// ValidateVersion checks whether the specified spec version is valid.
// In addition to checking whether the spec version is in the set of known versions, 
// the spec is inspected to determine whether the features used are available in specified
// version.
```

specs-go/version.go
```diff
@@ -60,13 +58,28 @@ var validSpecVersions = requiredVersionMap{
...
+func ValidateVersion(spec *Spec) error {
+	if !validSpecVersions.isValidVersion(spec.Version) {
+		return fmt.Errorf("invalid version %q", spec.Version)
+	}
+	minVersion, err := MinimumRequiredVersion(spec)
+	if err != nil {
+		return fmt.Errorf("could not determine minimum required version: %v", err)
```
elezar, line 68: nit: 
```suggestion
		return fmt.Errorf("could not determine minimum required version: %w", err)
```

specs-go/version.go
```diff
@@ -60,13 +58,28 @@ var validSpecVersions = requiredVersionMap{
...
+
 // MinimumRequiredVersion determines the minimum spec version for the input spec.
-func MinimumRequiredVersion(spec *cdi.Spec) (string, error) {
+func MinimumRequiredVersion(spec *Spec) (string, error) {
 	minVersion := validSpecVersions.requiredVersion(spec)
 	return minVersion.String(), nil
 }
 
-// version represents a semantic version string
+// Version represents a semantic version string
```
elezar, line 82:
```suggestion
// version represents a semantic version string
```

elezar (commented):
Thanks @bart0sh. It's looking a lot cleaner now.

I have some additional comments, but these are mostly limited to docstrings.

specs-go/version.go
elezar: This type was not public. See https://github.com/cncf-tags/container-device-interface/blob/e66544063aa7760c4ea6330ce9e6c757f8e61df2/pkg/cdi/version.go#L70

pkg/cdi/spec.go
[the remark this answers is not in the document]
elezar: This looks good now. Thanks.

We can make a decision later as to whether we want to add functions to the `Spec` type.

pkg/cdi/spec_test.go
bart0sh: This would add a bunch of dependencies to specs-go/go.mod/sum, which would make it harder to import into k/k

specs-go/version.go
bart0sh: ah, ok. thanks for pointing out.

pkg/cdi/spec_test.go
elezar: That is true. Let's leave it where it is then.

pkg/cdi/qualified-device.go
```diff
@@ -66,19 +66,6 @@ func ParseDevice(device string) (string, string, string) {
...
 
-// ParseQualifier splits a device qualifier into vendor and class.
-// The syntax for a device qualifier is
-//
-//	"<vendor>/<class>"
-//
-// If parsing fails, an empty vendor and the class set to the
-// verbatim input is returned.
-//
-// Deprecated: use parser.ParseQualifier instead
```
elezar, line 77: I think we should remove all deprecated functions with a single change.

Do we want to do that in a PR "before" this one, or as a follow-up?

specs-go/parser.go
```diff
@@ -0,0 +1,39 @@
...
+	parts := strings.SplitN(kind, "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", kind
+	}
+	return parts[0], parts[1]
+}
+
+// IsLetter reports whether the rune is a letter.
+func IsLetter(c rune) bool {
```
elezar, line 37: One thought on this. I don't want `IsLetter` to be part of the public API.

I can see use moving the `Parse` functions into the `specs` package and making these private. 

Do we want to mark this as deprecated to properly signal that this should not be considered "stable"?

elezar: The other option is to duplicate this functionality for now and remove it from the `parser` API later.

specs-go/parser.go
```diff
@@ -0,0 +1,39 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package specs
+
+import "strings"
+
+// ParseQualifier splits a device qualifier into vendor and class.
```
elezar, line 21: It's out of scope for this PR, but we may want to be stricter on our language here. In the spec the entity we're parsing here is the `Kind`, but this public function refers to it as a qualifier. Since we're in the `specs` package here, do we want to rename this to `SplitKind(string)` instead?

pkg/cdi/qualified-device.go
bart0sh: I'd prefer to do it in a follow-up PR.

specs-go/parser.go
bart0sh: Generally it makes sense to me. I'd propose to rename it to `ParseKind` if you don't mind. And I agree, it's out of scope of this PR.

pkg/cdi/qualified-device.go
elezar: Yes, a follow-up is better. Please re-add the removed code in that case.

pkg/cdi/spec.go
```diff
@@ -105,7 +106,7 @@ func newSpec(raw *cdi.Spec, path string, priority int) (*Spec, error) {
 		spec.path += defaultSpecExt
 	}
 
-	spec.vendor, spec.class = ParseQualifier(spec.Kind)
+	spec.vendor, spec.class = parser.ParseQualifier(spec.Kind)
```
elezar, line 109: Let's revert this change and do this as part of the follow-up to remove the deprecated functions.

specs-go/parser.go
elezar: OK. Let's do this as a follow-up then.

pkg/cdi/spec.go
```diff
@@ -328,7 +317,7 @@ func GenerateTransientSpecName(vendor, class, transientID string) string {
 // the Spec does not contain a valid vendor or class, it returns
 // an empty name and a non-nil error.
 func GenerateNameForSpec(raw *cdi.Spec) (string, error) {
-	vendor, class := ParseQualifier(raw.Kind)
+	vendor, class := parser.ParseQualifier(raw.Kind)
```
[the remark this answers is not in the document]
bart0sh: I think it's better to keep using the `parser` package.

elezar (approved):
Thanks for this @bart0sh.

I think the last comments I have are minor and can be addressed in follow-ups.

specs-go/parser.go
bart0sh: I decided to duplicate it and made all functions except of `ValidateVersion` and `MinimumRequiredVersion` private. PTAL.

specs-go/version.go
```diff
@@ -81,16 +97,16 @@ func (v version) String() string {
 }
 
 // IsGreaterThan checks with a version is greater than the specified version.
-func (v version) IsGreaterThan(o version) bool {
+func (v version) isGreaterThan(o version) bool {
```
elezar, line 100: Since the type is private we should not have to do this. If we do, please also update the docstring.

bart0sh: I'd prefer to explicitly set them public to avoid possible confusions. Updated the docstrings.

elezar: Should be good to merge as soon as you push this.

bart0sh: pushed. Please, merge, thanks!

PR #215, opened 2024-07-29
Document fields in spec to indicate when these were added.

bart0sh:
partial fix for #206

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -179,11 +179,13 @@ The keywords "must", "must not", "required", "shall", "shall not", "should", "sh
 * `kind` (string, REQUIRED) field specifies a label which uniquely identifies the device vendor.
   It can be used to disambiguate the vendor that matches a device, e.g: `docker/podman run --device vendor.com/device=foo ...`.
     * The `kind` label has two segments: a prefix and a name, separated by a slash (/).
-    * The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between.
+    * The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between. Dots (.) are supported from v0.6.0.
     * The prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).
     * Examples (not an exhaustive list):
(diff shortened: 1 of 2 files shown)


bart0sh:
@elezar @klihub PTAL

klihub (approved):

SPEC.md
```diff
@@ -179,12 +179,14 @@ The keywords "must", "must not", "required", "shall", "shall not", "should", "sh
 * `kind` (string, REQUIRED) field specifies a label which uniquely identifies the device vendor.
   It can be used to disambiguate the vendor that matches a device, e.g: `docker/podman run --device vendor.com/device=foo ...`.
     * The `kind` label has two segments: a prefix and a name, separated by a slash (/).
-    * The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between.
+    * The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between. Dots (.) are supported from v0.6.0.
```
elezar, line 182: Question: Do we also want to update the docs in the relevant spec fields?

elezar (commented):
For cases where fields were added, would it make sense to specify the version closer to the field name itself. e.g.: `Annotations` (string, OPTIONAL, v0.6.0)?

SPEC.md
bart0sh: Added comments with version info. PTAL.

bart0sh:
> For cases where fields were added, would it make sense to specify the version closer to the field name itself. e.g.: `Annotations` (string, OPTIONAL, v0.6.0)?

I'd leave it as it is. "Added in v.6.0" sounds more informative to me.

specs-go/config.go
```diff
@@ -19,6 +20,7 @@ type Spec struct {
 type Device struct {
 	Name string `json:"name"`
 	// Annotations add meta information per device. Note these are CDI-specific and do not affect container metadata.
+	// Added in v0.6.0.
```
elezar, line 23: One thought that I had was whether we wanted to add golang field tags so that we could parse these programatically? This could be done as a follow-up though.

specs-go/config.go
```diff
@@ -10,6 +10,7 @@ type Spec struct {
 	Version string `json:"cdiVersion"`
 	Kind    string `json:"kind"`
```
elezar, line 11: Do we want to call out the allowed character change for the class in a docstring?

PR #219, opened 2024-08-08
Remove deprecated registry API

bart0sh:


bart0sh:
This one is not as straightforward as the previous one, so @klihub please take a look. I'm not sure replacing `GetRegistry` with `GetDefaultCache` is correct approach.

bart0sh:
@klihub 
> But one thing I already noticed will need updating is our godoc in pkg/cdi/doc.go.

Thank you for pointing this out! I've updated the docs and code comments.

bart0sh:
@elezar @klihub Can you review this PR, please?

cmd/cdi/cmd/root.go
```diff
@@ -35,13 +35,13 @@ var (
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
 	Use:   "cdi",
-	Short: "Inspect and interact with the CDI Registry",
+	Short: "Inspect and interact with the CDI Cache",
```
elezar, line 38: nit: Should this be `cache` instead of `Cache`?

cmd/cdi/cmd/root.go
```diff
@@ -68,10 +68,11 @@ func initSpecDirs() {
...
 	if len(specDirs) > 0 {
-		cdi.GetRegistry(
-			cdi.WithSpecDirs(specDirs...),
-		)
-		if len(cdi.GetRegistry().GetErrors()) > 0 {
+		cache := cdi.GetDefaultCache()
+		if err := cache.Configure(cdi.WithSpecDirs(specDirs...)); err != nil {
+			fmt.Printf("failed to configure CDI cache: %v\n", err)
+		}
+		if len(cache.GetErrors()) > 0 {
```
elezar, line 75:
```suggestion
		cache, err := cdi.NewCache(
			cdi.WithSpecDirs(specDirs...),
		)
		if err != nil {
			fmt.Printf("failed to configure CDI cache: %v\n", err)
		}
		if len(cache.GetErrors()) > 0 {
```

cmd/cdi/cmd/root.go
```diff
@@ -68,10 +68,11 @@ func initSpecDirs() {
 	cdi.SetSpecValidator(validate.WithSchema(s))
 
 	if len(specDirs) > 0 {
-		cdi.GetRegistry(
-			cdi.WithSpecDirs(specDirs...),
-		)
-		if len(cdi.GetRegistry().GetErrors()) > 0 {
+		cache := cdi.GetDefaultCache()
+		if err := cache.Configure(cdi.WithSpecDirs(specDirs...)); err != nil {
+			fmt.Printf("failed to configure CDI cache: %v\n", err)
```
elezar, line 73: Is printing an error sufficient?

pkg/cdi/doc.go
```diff
@@ -35,48 +35,35 @@
...
 // # Device Injection
 //
-// Using the Registry one can inject CDI devices into a container with code
+// Using the Cache one can inject CDI devices into a container with code
 // similar to the following snippet:
 //
-//	import (
-//	    "fmt"
-//	    "strings"
+//		import (
```
elezar, line 45: Is the whitespace required?

pkg/cdi/doc.go
```diff
@@ -99,40 +86,40 @@
 // CDI Spec cache is up to date before performing device injection.
 // A code snippet similar to the following accmplishes that:
 //
-//	import (
-//	    "fmt"
-//	    "strings"
+//		import (
```
elezar, line 89: Similar here.

pkg/cdi/doc.go
```diff
@@ -186,13 +173,13 @@
 //
 //	"fmt"
 //	...
-//	"tags.cncf.io/container-device-interface/specs-go"
+//	specs "tags.cncf.io/container-device-interface/specs-go"
```
elezar, line 176: Is the `specs` alias strictly needed?

pkg/cdi/regressions_test.go
```diff
@@ -148,12 +148,16 @@ func withCDI(t *testing.T, annotations map[string]string, cdiSpecDirs []string)
 			return nil
 		}
 
-		registry := GetRegistry(WithSpecDirs(cdiSpecDirs...))
-		if err = registry.Refresh(); err != nil {
-			t.Logf("CDI registry refresh failed: %v", err)
+		cache := GetDefaultCache()
+		if err := cache.Configure(WithSpecDirs(cdiSpecDirs...)); err != nil {
```
elezar, line 152:
```suggestion
		cache, err := NewCache(
			WithSpecDires(cdiSpecDirs...),
		)
		if err != nil {
```

elezar (changes requested):
Thanks @bart0sh.

I have some comments regarding docs. I also think that when it makes sense to do so, we should rather instantiate a `cache` instance explicitly instead of relying on the (global) default cache. Especially in cases where we are configuring the spec directories.

bart0sh:
@elezar @klihub All review comments should be addressed now, please take a look again, thanks!

pkg/cdi/doc.go
bart0sh: removed

pkg/cdi/doc.go
bart0sh: yeah, it's my editor reformatted those. Should be fixed now. Unfortunately with more reformatting.

cmd/cdi/cmd/validate.go
```diff
@@ -26,22 +26,23 @@ import (
...
-were reported by the registry.`,
+of the CDI cache. It exits with an exit status of 1 if any errors
+were reported by the cache.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		cdiErrors := cdi.GetRegistry().GetErrors()
+		cache := cdi.GetDefaultCache()
+		cdiErrors := cache.GetErrors()
 		if len(cdiErrors) == 0 {
-			fmt.Printf("No CDI Registry errors.\n")
+			fmt.Printf("No CDI Cache errors.\n")
```
klihub, line 41: Apparently the original code was inconsistent in this regard, but maybe we could fix it and should always use either lowercase `cache` or capitalized `Cache` in help messages and error/info strings in the `cdi` command. This PR already mostly uses lowercase, but this is an exception here.

cmd/cdi/cmd/validate.go
```diff
@@ -26,22 +26,23 @@ import (
...
 			return
 		}
 
-		fmt.Printf("CDI Registry has errors:\n")
+		fmt.Printf("CDI Cache has errors:\n")
```
klihub, line 45: nit: ditto lower vs. upper case here...

klihub (approved):
@bart0sh I have two very small nitpicks about help and info/error message consistency wrt. `cache` vs. `Cache` in messages. The original code was apparently inconsistent in this regard (no wonder, knowing the author), but this PR is already very close two fixing that with two exceptions.

Otherwise LGTM. Not a showstopper though, hence the approval.

cmd/cdi/cmd/validate.go
bart0sh: Thank you for the review! Should be fixed now.

cmd/cdi/cmd/validate.go
bart0sh: done

pkg/cdi/doc.go
```diff
@@ -106,17 +93,17 @@
...
 //
 //	    "tags.cncf.io/container-device-interface/pkg/cdi"
-//	    oci "github.com/opencontainers/runtime-spec/specs-go"
+//	    "github.com/opencontainers/runtime-spec/specs-go"
 //	)
 //
-//	func injectCDIDevices(spec *oci.Spec, devices []string) error {
-//	    registry := cdi.GetRegistry()
+//	func injectCDIDevices(spec *specs.Spec, devices []string) error {
+//		cache := cdi.GetDefaultCache()
```
elezar: nit: indent

elezar (approved):
I made some changes to the whitespace in doc.go.

Looks good otherwise.

pkg/cdi/doc.go
bart0sh: Thank you. Can you merge the PR?

PR #220, opened 2024-08-09
Release table: add forgotten v0.5.0 change

bart0sh:


SPEC.md
```diff
@@ -195,7 +196,7 @@ Note: For a CDI file to be valid, at least one entry must be specified in this a
   * `devices` (array of objects, REQUIRED) list of devices provided by the vendor.
     * `name` (string, REQUIRED), name of the device, can be used to refer to it when requesting a device.
       * Beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between.
```
elezar, line 198: It seems as if the addition should not be on the example line, maybe rather on the line above?

bart0sh: `e.g.` is usually used as a part of sentence, so I thought this is the case here. I've replaced `.` with `.`. I think this is better than 
```
* Beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between. Beginning with digits is supported from v0.5.0.
* e.g: `docker/podman run --device foo ...`.
```
as this way it's a bit confusing, at least to me.

bart0sh: Alternatively we can drop an example as it's not obvious how it's related to the explanation.
So, result would look like this:
```
* Beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (\_), dots (.), and alphanumerics between. Beginning with digits is supported from v0.5.0.
```

elezar: I think dropping the example may make sense. Let's do that.

bart0sh: done

elezar (approved):

PR #221, opened 2024-08-09
Remove specs-go/parsing APIs

bart0sh:
Got rid of using `parseQualifier` and `isLetter` APIs in version.go


specs-go/version.go
```diff
@@ -183,23 +183,20 @@ func requiresV060(spec *Spec) bool {
 	}
 
 	// The v0.6.0 spec allows dots "." in Kind name label (class)
-	vendor, class := parseQualifier(spec.Kind)
-	if vendor != "" {
-		if strings.ContainsRune(class, '.') {
-			return true
-		}
+	if !strings.Contains(spec.Kind, "/") {
```
elezar, line 186: There's a subtle difference here in that if there is no `/` the *class* is assumed to equal the Kind. and that would be checked for a `.`.

I don't think that's used in practice though.

elezar: OK. Wait, I've just seen the other conditional below.

elezar (approved):
LGTM

bart0sh:
@elezar Thank you for the fast review & merging!

PR #225, opened 2024-08-20
Add description of standard hooks

elezar:
This change adds a description of standard hooks to the CDI specification. These hooks can be used to update the LDCache in the container or to create symlinks.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -245,6 +245,53 @@ The `containerEdits` field has the following definition:
     * `enableMBM` (boolean, OPTIONAL) whether to enable memory bandwidth monitoring
   * `additionalGids` (array of uint32s, OPTIONAL) A list of additional group IDs to add with the container process. These values are added to the `user.additionalGids` field in the OCI runtime specification. Values of 0 are ignored. Added in v0.7.0.
 
+## Hooks
+
+Hooks allow CDI spec vendors to inject specific logic into various points of the
+container lifecycle. These are typically mapped to [OCI runtime hooks](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks) and
+are typically used when behaviour depends on the container contents in some way.
+
+The following named hooks are defined by the CDI specification:
+* `update-ldcache`
+* `create-symlinks`
+
+Note that although their intent is defined as part of the CDI specification, their
+implementation and distribution is left to vendors. Note that these hooks are both
+`createContainer` hooks, meaning that in the context of OCI-compliant runtimes,
+the hook path is resolved in the runtime namespace, while it is executed in the
+container namespace.
+
+### update-ldcache
+
+The `update-ldcache` hook is a `createContainer` hook that is used to ensure that
+the ldcache in a container is updated to include any injected libraries.
+
+Assuming the following hook syntax:
+```shell
+[command-prefix] update-ldcache [--folder folder1] [--folder folder2]
+```
+executing the `update-ldcache` hook will:
+1. ensure that libraries in requested folders (`folder1`, `folder2`) are added to the `ldcache` in the container with the correct priority.
+2. create the relevant `.so.SONAME` symlinks in the container.
+
+Note that if updating the ldcache in the container is not applicable, this is skipped, but the symlinks are still created.
+
+### create-symlinks
+
+The `create-symlinks` hook is a `createContainer` hooks that is used to ensure that
+required symlinks exist in a container. Typically these symlinks point to injected
+libraries or executables.
+
+Assuming the following hook syntax:
+```shell
+[command-prefix] create-symlinks [--link target::link-path]
+```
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -245,6 +245,53 @@ The `containerEdits` field has the following definition:
...
+
+### create-symlinks
+
+The `create-symlinks` hook is a `createContainer` hooks that is used to ensure that
+required symlinks exist in a container. Typically these symlinks point to injected
+libraries or executables.
+
+Assuming the following hook syntax:
+```shell
+[command-prefix] create-symlinks [--link target::link-path]
```
elezar, line 287: I'm not 100% happy with the `target::link-path` format here. Are there other suggestions on how we would do this?

Something like `--link source=foo,target=bar` instead?

klihub: Cannot we use an ln:ish syntax of `--link $existing_target $new_link_name` ?

reidpr: I agree with @elezar's initial concern. The way I'd frame it is, because the hook arguments is a JSON list, more syntax within those list elements introduces a new DSL.

The `foo::bar` notation is new to me. I'm guessing it's more standard in golang world, but in Python or C it would require a second stage of argument parsing.

My suggestions would be:

* Adopt a general principle of one value per argument (i.e., no `::` operator).
* Specifically, use @klihub's suggestion of separate target and source arguments.
* Remove `--link` because it's redundant with the hook definition. This would also make argument parsing in Python or C easier because the target and source become positional arguments rather than options for `--link`. (Also yes, an argument that takes multiple options is hard to parse in my world too.) `nvidia-ctk-hook` could simply ignore `--link` to be backwards compatible.
* Make at least one target and one source required.
* Clarify whether the hook takes multiple target/source pairs or just one.

reidpr (commented):
Thanks for requesting my opinion.

I think this is a big step forward for the standard.

It does still seem roundabout to me. That is, with the introduction of standard hooks, we now have prescribed behavior, which is great, but it's indirected through the constraints and syntax of command-line programs. To me it would be simpler to just introduce new `updateLDCache` and `createSymlinks` fields.

Further suggestions in-line. I hope this is helpful.

SPEC.md
```diff
@@ -245,6 +245,53 @@ The `containerEdits` field has the following definition:
...
 
+## Hooks
+
+Hooks allow CDI spec vendors to inject specific logic into various points of the
+container lifecycle. These are typically mapped to [OCI runtime hooks](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks) and
+are typically used when behaviour depends on the container contents in some way.
+
+The following named hooks are defined by the CDI specification:
+* `update-ldcache`
+* `create-symlinks`
```
klihub, line 256: @elezar One question related to this... Is the idea with this document addendum, to describing de-facto hooks, including their command line options and expected behavior ? Also, is the `[command prefix] $named_hook` syntax to allow for an implementation, where a single binary implements all the de-facto hooks (and an extra 1st $hook_name argument is used to pick which hook to run from that binary) ?

If it is so, and the CDI specification provides such a detailed description of these de-facto hooks then wouldn't it make also sense to provide a default implementation for these hooks ?

SPEC.md
elezar: One thing that's not in the syntax above is that `--link` can be repeated meaning that one can create multiple links with a single invocation of the `create-symlink` command.

bart0sh: @elezar How would it look in the CDI file and how it would differ from the currently available approach:
```
"hooks": [
    {"startContainer": {"path": "/usr/bin/ln", "args": ["/src", "/tgt"] }
]

```

SPEC.md
```diff
@@ -250,6 +250,106 @@ The `containerEdits` field has the following definition:
...
+### create-symlinks
+
+The `create-symlinks` hook is a `createContainer` hook that is used to ensure that
+required symlinks exist in a container. Typically these symlinks point to injected
+libraries or executables.
+
+The hooks will have the following YAML definition:
+```yaml
+  hooks:
+  - path: <comnand-prefix>
```
klihub, line 279: typo: comnand-prefix should be command-prefix

SPEC.md
```diff
@@ -250,6 +250,106 @@ The `containerEdits` field has the following definition:
...
+
+The `update-ldcache` hook is a `createContainer` hook that is used to ensure that
+the ldcache in a container is updated to include any injected libraries. This
+ensures that a user does not have to update environment variables such as `LD_LIBRARY_PATH`
+themselves. The hook also creates `SONAME` symlinks for any injected libraries.
+
+The hooks will have the following YAML definition:
+```yaml
+  hooks:
+  - path: <comnand-prefix>
```
klihub, line 321: typo: ditto here

klihub (changes requested):
I have two typo-nits that should be fixed.

Issue #227, opened 2024-08-27
Inconsistent naming of the specs-go module


bart0sh:
```
$ git grep 'container-device-interface/specs-go"'
pkg/cdi/cache.go:       cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/cache_test.go:  cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/container-edits.go:     "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/container-edits_test.go:        cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/device.go:      cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/device_test.go: cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/doc.go://       "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/doc.go://       "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/registry.go:    cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/spec.go:        cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/spec_test.go:   cdi "tags.cncf.io/container-device-interface/specs-go"
pkg/cdi/validate/schema.go:     raw "tags.cncf.io/container-device-interface/specs-go"
```

I'd propose to replace all of that with unnamed import `"tags.cncf.io/container-device-interface/specs-go"`

klihub:
Isn't that consistent naming ? I mean everywhere within pkg/cdi, we call specs-go cdi, so the spec reference reads cdi.Spec, not specs.Spec. Similarly we call opencontainers/runtime-spec/spec-go oci, so we have then oci.Spec in the code.

Anyway you need to import one of them with a renamed alias, since you can't call both of them in the same module specs. So then probably you would have specs.Spec instead of cdi.Spec and maybe oci.Spec or ocispec.Spec for the oci Spec. Personally I consider both of these less readable than the current implementation choice.

klihub:
Ah ok. In pkg/container-edits.go we're inconsistent. If the current and above described reasoning is acceptable, then at least that should be changed to be consistent with the rest within pkg/cdi. And then I'd change raw to cdi in pkg/cdi/validate/schema.go as well.

klihub:
[...]

PR #228, opened 2024-08-28
pkg/cdi: name cdi specs-go imports consistently.

klihub:
Addresses inconsistent naming of imports, as reported in #227.

### DIFF
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -26,7 +26,7 @@ import (
 
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	ocigen "github.com/opencontainers/runtime-tools/generate"
-	"tags.cncf.io/container-device-interface/specs-go"
+	cdi "tags.cncf.io/container-device-interface/specs-go"
 )
 
 const (
@@ -64,7 +64,7 @@ var (
 // to all OCI Specs where at least one devices from the CDI Spec
 // is injected.
 type ContainerEdits struct {
-	*specs.ContainerEdits
+	*cdi.ContainerEdits
 }
 
 // Apply edits to the given OCI Spec. Updates the OCI Spec in place.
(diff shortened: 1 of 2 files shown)


pkg/cdi/container-edits.go
```diff
@@ -26,7 +26,7 @@ import (
 
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	ocigen "github.com/opencontainers/runtime-tools/generate"
-	"tags.cncf.io/container-device-interface/specs-go"
+	cdi "tags.cncf.io/container-device-interface/specs-go"
```
bart0sh, line 29: Would it be better to import tags.cncf.io/container-device-interface/specs-go as is to avoid confusion with pkg/cdi?

bart0sh: I thought about it. renaming it to cdispecs would trigger renaming oci -> ocispecs for consistency, so I'm still for letting it be specs, i.e. using not named import, unless it triggers linting failures.

elezar: OK. `specs` is fine.

klihub: > Would it be better to import tags.cncf.io/container-device-interface/specs-go as is to avoid confusion with pkg/cdi?

No, not in my opinion, since this is the `pkg/cdi` package itself so it never needs to reference itself `externally`. IMO/for me the most readable alternative is to reference the OCI Spec as `oci.Spec` and the CDI Spec as `cdi.Spec` in the code, with the next best option being `ocispec.Spec` and `cdispec.Spec` and that's why I chose to do it that way.

But if I in a minority with this opinion with some other consensus on what's best, just let me 
guys know what is the consensus and I'll update this PR accordingly.

bart0sh: That makes sense to me.
/lgtm

if external project imports both the cdi and the specs-go packages they still can import them as is, i.e. as cdi and specs.

bart0sh (approved):

elezar (approved):

PR #232, opened 2024-10-15
fixes: enable golanci-lint in CI, fix complaints.

klihub:
This patch set
- adds a local [golang]ci-lint Makefile target and hooks it in to pr-checks
- addresses golangci-lint complaints
- adds CI golangci-lint workflow

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,27 @@
+name: CI
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.61.0
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -116,7 +116,7 @@ func (c *Cache) configure(options ...Option) {
 		c.watch.setup(c.specDirs, c.dirErrors)
 		c.watch.start(&c.Mutex, c.refresh, c.dirErrors)
 	}
-	c.refresh()
+	_ = c.refresh() // we record but ignore errors
 }
 
 // Refresh rescans the CDI Spec directories and refreshes the Cache.
@@ -233,7 +233,7 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 	c.Lock()
 	defer c.Unlock()
 
-	c.refreshIfRequired(false)
+	_, _ = c.refreshIfRequired(false) // we record but ignore errors
 
 	edits := &ContainerEdits{}
 	specs := map[*Spec]struct{}{}
@@ -340,7 +340,7 @@ func (c *Cache) GetDevice(device string) *Device {
 	c.Lock()
 	defer c.Unlock()
 
-	c.refreshIfRequired(false)
+	_, _ = c.refreshIfRequired(false) // we record but ignore errors
 
 	return c.devices[device]
 }
@@ -352,7 +352,7 @@ func (c *Cache) ListDevices() []string {
 	c.Lock()
 	defer c.Unlock()
 
-	c.refreshIfRequired(false)
+	_, _ = c.refreshIfRequired(false) // we record but ignore errors
 
 	for name := range c.devices {
 		devices = append(devices, name)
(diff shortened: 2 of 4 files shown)


pkg/cdi/default-cache_test.go
```diff
@@ -115,7 +115,7 @@ devices:
 					filepath.Join(dir, "run"),
 				),
 			}
-			Configure(opts...)
+			_ = Configure(opts...) // always nil now, should we update and break API ?
```
bart0sh, line 118: Would it make sense to check the error and fail a test if it's not nil? If/when Configure starts to return an error, we don't need to modify the tests.

elezar, line 118: I think we said that we could foresee a future where the configure call MAY return an error.

pkg/cdi/cache.go
```diff
@@ -340,7 +340,7 @@ func (c *Cache) GetDevice(device string) *Device {
 	c.Lock()
 	defer c.Unlock()
 
-	c.refreshIfRequired(false)
+	_, _ = c.refreshIfRequired(false) // we record but ignore errors
```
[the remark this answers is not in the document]
elezar: Are there places where we actually check the error return values for the internal calls?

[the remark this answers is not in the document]
klihub: > Can we call `c.Refresh(`) here?

No, I don't think so. `Refresh()` is a locked `refreshIfRequired()` which always returns collected errors if it ended up doing a refresh. So it would be the same if it did not deadlock itself.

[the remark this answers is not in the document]
klihub: > Are there places where we actually check the error return values for the internal calls?

We check and propagate errors back on the caller only on the externally triggered calls (`Refresh()`), not on the internally triggered ones. But we handle both with a single code path.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.61.0
```
elezar, line 27: Do we want to specify the version, or should we just use the "default"?

pkg/cdi/cache.go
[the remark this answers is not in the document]
elezar: Was just able to look through the code. Thanks for confirming.

elezar (approved):
I think this looks good @klihub.

One thought is that for cases where we do record but ignore the errors, we should update the public API documentation to indiate that `GetErrors()` can be queried to return errors that were raised when refreshing the cache. 

Out of scope: we could also look at further segmenting the API to functions that are used by our container runtime consumers (basically Cache instantiation and InjectDevices) so that we have more flexibility with other utility functions such as getting lists of devices.

.github/workflows/golangci-lint.yml
klihub: I'm not sure what would make the most sense here. I simply took the provided sample workflow as a basis for this, then bumped the version to the latest release, which I then also used locally for linting.

Looking at the action implementation, it is stated that we could use an explicit 'latest' here. But TBH, I'm not too keen on a setup where we are not in control so without us changing anything, things already merged might start getting flagged in new PRs as linting errors.

Also looking at other projects, most of them seem to lock the version explicitly to one of their choice, which they then keep gradually updating. If we really wanted to go with an uncontrolled rolling version, then we'd need to add a nightly run, to catch extra-repo golangci-lint changes breaking things for us...

klihub:
> I think this looks good @klihub.
> 
> One thought is that for cases where we do record but ignore the errors, we should update the public API documentation to indiate that `GetErrors()` can be queried to return errors that were raised when refreshing the cache.

I updated the doc strings of the relevant Cache public API functions, with a brief mention a possible implicit cache refresh and how to obtain any errors encountered it. PTAL.

.github/workflows/golangci-lint.yml
bart0sh: I agree with @klihub here. To have a reproducible builds (tests in this case) we need to lock lint version.

pkg/cdi/default-cache_test.go
```diff
@@ -115,7 +115,8 @@ devices:
 					filepath.Join(dir, "run"),
 				),
 			}
-			Configure(opts...)
+			err = Configure(opts...)
+			require.Nil(t, err, "unexpected Configure() error")
```
bart0sh, line 119: `require.NoError` would be more explicit here in my opinion.

pkg/cdi/cache.go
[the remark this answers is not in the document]
klihub: > @klihub
> 
> > Refresh() is a locked refreshIfRequired() which always returns collected errors if it ended up doing a refresh. So it would be the same but would deadlock itself.
> 
> Sorry for not being clear. I actually meant using `Refresh` and remove locking on lines 340 and 341. Are those locks needed for anything except locking `refreshIfRequired`?

The locks are for providing atomicity between refresh scanning Spec directories and updates the Specs/the Cache wrt. any other externally triggered access to the cache accessing the same data. IOW, you either get access to the data in the cache before a refresh cycle or after, but not while a refresh is happening.

pkg/cdi/default-cache_test.go
klihub: Thanks for spotting it. Updated accordingly.

bart0sh (approved):
/lgtm

PR #233, opened 2024-10-15
Add producer API to write specs

elezar:
This change proposes a "producer" API for generating and writing specs.

The more general `cdi` package imports this package when writing specs.


pkg/cdi/producer/producer.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := sp.save(filename, !p.failIfExists); err != nil {
+		return "", err
+	}
+
+	return filename, nil
+}
+
+// Validate performs a validation on a CDI spec using the configured validator.
+// If no validator is configured, the spec is considered unconditionaly valid.
```
klihub, line 72: codespell nit.

pkg/cdi/producer/options.go
```diff
@@ -0,0 +1,59 @@
...
+			p.format = format
+		default:
+			return fmt.Errorf("invalid CDI spec format %v", format)
+		}
+		return nil
+	}
+}
+
+// WithSpecValidator sets a validator to be used when writing an output spec.
+func WithSpecValidator(v specValidator) Option {
```
klihub, line 42: I would expose the `specValidator` interface by calling it `SpecValidator`.

pkg/cdi/producer/api.go
```diff
@@ -0,0 +1,36 @@
...
+	DefaultSpecFormat = SpecFormatYAML
+
+	// SpecFormatJSON defines a CDI spec formatted as JSON.
+	SpecFormatJSON = specFormat(".json")
+	// SpecFormatYAML defines a CDI spec formatted as YAML.
+	SpecFormatYAML = specFormat(".yaml")
+)
+
+// a specValidator is used to validate a CDI spec.
+type specValidator interface {
```
klihub, line 34: nit: Shouldn't we export this interface, too (even if it is so trivial) ?

The package-provided `Default` and `Disabled` implementations are exported (and are supposed to be referenced directly as such, and not for instance indirectly by a name, etc.). I am guessing that we want to allow folks to plug in their own validation logic. Since (assuming those folks use the stock c-d-i package also for eventual interpretation/injection) custom validators should always be stricter than the default one (to ensure injection by the same package wouldn't fail). That suggests that most custom validators will be a composition of the default validator + some extra custom restrictions.

Now, *it is already possible* to do composition using the default validator, but if my aforementioned assumptions are true, it feels strange to not expose the interface itself.

pkg/cdi/producer/producer.go
```diff
@@ -0,0 +1,86 @@
...
+// A Producer defines a structure for outputting CDI specifications.
+type Producer struct {
+	format       specFormat
+	failIfExists bool
+}
+
+// New creates a new producer with the supplied options.
+func New(opts ...Option) (*Producer, error) {
+	p := &Producer{
+		format: DefaultSpecFormat,
```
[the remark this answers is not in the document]
klihub: > Do we know if this would be an allowed dependency in the k8s case ?

Just based on a quick git grep, it is an indirect dependency hack hack/tools, but not k8s itself. So I'm guessing... inconclusive.

elezar:
@klihub I simplified this PR a bit to start with. If you're ok with the general direction, we can get this in and then move some of the other functionalty (annotations, device name generation) to this package as well.

I would also like to add functionality to allow the minimum verison to be determined at save, but would rather do that as a follow up.

pkg/cdi/producer/producer.go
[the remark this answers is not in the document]
elezar: It seems as if it used to be a top-level import but was recently removed.

pkg/cdi/producer/api.go
elezar: My thinking was that one could plug in the schema validators here as well. I had some local changes that would expose those as interfaces instead of as raw function pointers.

I'm happy to expose the interface as public though to make the intent clearer and allow users to compose these as required.

pkg/cdi/producer/api.go
```diff
@@ -0,0 +1,37 @@
...
+	// SpecFormatJSON defines a CDI spec formatted as JSON.
+	SpecFormatJSON = SpecFormat(".json")
+	// SpecFormatYAML defines a CDI spec formatted as YAML.
+	SpecFormatYAML = SpecFormat(".yaml")
+)
+
+// A SpecValidator is used to validate a CDI spec.
+type SpecValidator interface {
+	Validate(*cdi.Spec) error
+}
```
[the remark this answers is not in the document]
elezar: I don't want to add unnecessary things to the `specs-go` package. I think we made the call that code to determine the minimum version was close enough to the spec to implement there, but I think an operation on the spec such as validation is not.

Note that due to the nature of go interfaces, this type need not be exported at all -- although it may be useful if it is.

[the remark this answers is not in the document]
bart0sh: Sorry for not being clear with this. I didn't mean moving only SpecValidator interface. I was talking about moving the whole validator package there. The idea was that it doesn't logically belong to producer. Having it here looks a bit confusing to me. 

There is at least one reason not to move it to the spec module though. If it brings a lot of additional external dependencies to the spec package, we should probably avoid doing this. 

@klihub @elezar WDYT?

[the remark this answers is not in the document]
elezar: I would rather suggest that we move the version specifics out of `specs-go` implement them as part of the "producer" package. Ideally we would implement this as a submodule as well so that we can properly separate the dependencies.

In the end we will have `specs-go` defining the actual data structures and helpers around this -- including validation. Implemented in other packages / submodules.

pkg/cdi/producer/producer.go
```diff
@@ -0,0 +1,188 @@
...
+	filename = p.normalizeFilename(filename)
+	format := p.specFormatFromFilename(filename)
+	// If the currently configured format doesn't match the expected format for
+	// the filename, we create a new
+	if p.specFormat != format {
+		pWithFormat, err := cloneWithOptions(p, WithSpecFormat(format))
+		if err != nil {
+			return "", err
+		}
+		return pWithFormat.Save(filename)
```
klihub, line 87: nit: again, this it not a biggie, but it feels a bit weird, having to 'clone' the producer just to change the format... and clone is only used for this.

pkg/cdi/producer/api.go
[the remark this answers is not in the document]
klihub: > Sorry for not being clear with this. I didn't mean moving only SpecValidator interface. I was talking about moving the whole validator package there. The idea was that it doesn't logically belong to producer. Having it here looks a bit confusing to me.


> There is at least one reason not to move it to the spec module though. If it brings a lot of additional external dependencies to the spec package, we should probably avoid doing this.
> 
> @klihub @elezar WDYT?

Yes, I had the same gut feeling. As long as the producer and the consumer were under a single package, having the validator as a subpackage thereof felt like a logical choice. Now if we split them, it does not feel so natural to have it under either... so maybe a `pkg/validator`. This can be adjusted with subsequent PRs, too.

pkg/cdi/producer/api.go
klihub: Having the ability to plug in additional schema-based validation here feels like a good idea.

pkg/cdi/producer/options.go
```diff
@@ -0,0 +1,75 @@
...
+
+type options struct {
+	specFormat    SpecFormat
+	specValidator SpecValidator
+	overwrite     bool
+	permissions   fs.FileMode
+}
+
+// WithSpecFormat sets the output format of a CDI specification.
+func WithSpecFormat(format SpecFormat) Option {
```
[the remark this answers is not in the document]
elezar: Note that the type also exposes a `WriteTo(io.Writer)` function which is where this spec format is used. As you note, the `Save(string)` function does override the format to be consistent with the filename.

Happy to change the method name though.

pkg/cdi/producer/producer.go
elezar: An initial implementation that I had was to have a separate type to do the writing / saving and to construct this when `Save` is called. Let me look at it a bit again and see if I can come up with a better abstraction.

pkg/cdi/producer/api.go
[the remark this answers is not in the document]
elezar: I was actually wondering whether we shouldn't make this split at a module level so as to better manage dependencies. Meaning that we'd have:
```
tags.cncf.io/container-device-interface/api/v1/consumer
tags.cncf.io/container-device-interface/api/v1/producer
```
as top-level go submodules.

For the producer specifically, we would start out with:
```
tags.cncf.io/container-device-interface/api/v1/producer/cdi
```
with the types in this package.

We could keep the existing `pkg/cdi` package to provide backward compatibility.

A client that produces specs would then be something like:
```
package main
import (
  "tags.cncf.io/container-device-interface/api/v1/producer/cdi"
)

func main() {
    raw := generateRawCDISpec()
    saver, _ := cdi.New(raw, cdi.WithPermissions(0644))
    err := saver.Save("/var/run/cdi/example.json)
}
```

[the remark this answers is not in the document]
klihub: > A client that produces specs would then be something like:
> 
> ```
> package main
> import (
>   "tags.cncf.io/container-device-interface/api/v1/producer/cdi"
> )
> 
> func main() {
>     raw := generateRawCDISpec()
>     saver, _ := cdi.New(raw, cdi.WithPermissions(0644))
>     err := saver.Save("/var/run/cdi/example.json)
> }
> ```

As abstractions go, it feels a bit artificial to have a 1-to-1 association between a producer and a Spec, IOW to essentially make `SpecProducer` a simple Spec wrapper. I would expect a producer to produce one or more Spec files which then typically share the same file format and permissions. So I would make those options, and I'd rather pass the raw spec to be written to the Save or some similar function and not to the constructor itself.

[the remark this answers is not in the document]
elezar: That's a fair point. One question in this regard, let's say that the producer is configured to set the minimum version of the spec on save is it OK to modify the spec in-place?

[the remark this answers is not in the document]
bart0sh: > For the producer specifically, we would start out with: tags.cncf.io/container-device-interface/api/v1/producer/cdi

It'd look a bit inconsistent that producer and consumer APIs have a bit different paths. I'd propose to use
 tags.cncf.io/container-device-interface/api/v1/producer and tags.cncf.io/container-device-interface/api/v1/consumer as [you've proposed above](https://github.com/cncf-tags/container-device-interface/pull/233#discussion_r1817723622).

bart0sh:
@elezar @klihub I thought that we agreed to introduce validator package as a separate PR, no?

klihub:
> @elezar @klihub I thought that we agreed to introduce validator package as a separate PR, no?

That was also my understanding.

api/producer/api.go
```diff
@@ -0,0 +1,34 @@
...
+	// SpecFormatJSON defines a CDI spec formatted as JSON.
+	SpecFormatJSON = SpecFormat(".json")
+	// SpecFormatYAML defines a CDI spec formatted as YAML.
+	SpecFormatYAML = SpecFormat(".yaml")
+)
+
+// A SpecValidator is used to validate a CDI spec.
+type SpecValidator interface {
+	Validate(*cdi.Spec) error
+}
```
elezar, line 34: This could strictly speaking be removed.

pkg/cdi/cache.go
```diff
@@ -283,28 +284,35 @@ func (c *Cache) highestPrioritySpecDir() (string, int) {
...
 		err     error
 	)
-
 	specDir, prio = c.highestPrioritySpecDir()
 	if specDir == "" {
 		return errors.New("no Spec directories to write to")
 	}
 
-	path = filepath.Join(specDir, name)
-	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
```
elezar, line 298: This was removed since the same check is already done in `newSpec`.

elezar:
@bart0sh @klihub sorry for the delay here. I have pulled the validation specifics into #247.

pkg/cdi/spec.go
```diff
@@ -259,7 +207,7 @@ func SetSpecValidator(fn func(*cdi.Spec) error) {
 	specValidator = fn
 }
 
-// validateSpec validates the Spec using the extneral validator.
+// validateSpec validates the Spec using the extneral validation.
```
klihub, line 210: nit: old typo `extneral` retained.

api/producer/options.go
```diff
@@ -0,0 +1,61 @@
...
+
+package producer
+
+import (
+	"fmt"
+	"io/fs"
+)
+
+// An Option defines a functional option for constructing a producer.
+type Option func(*options) error
```
bart0sh, line 25: If this is an external API then `options` should be also public, no?

bart0sh: I take it back. It looks like options could only be set by With* public calls.

api/producer/api.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	// DefaultSpecFormat defines the default encoding used to write CDI specs.
+	DefaultSpecFormat = SpecFormatYAML
+
+	// SpecFormatJSON defines a CDI spec formatted as JSON.
+	SpecFormatJSON = SpecFormat(".json")
+	// SpecFormatYAML defines a CDI spec formatted as YAML.
+	SpecFormatYAML = SpecFormat(".yaml")
+)
```
bart0sh, line 29: Should these constants be moved to specs-go? They seem to logically belong to spec and don't introduce any new deps.

api/producer/api.go
bart0sh: Or moved to the https://github.com/cncf-tags/container-device-interface/pull/247

api/producer/set-minimum-version.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+
+	cdi "tags.cncf.io/container-device-interface/specs-go"
+)
+
+type setMinimumRequiredVersion struct{}
+
+// Transform detects the minimum required version required for the specified
+// spec and sets the version field accordingly.
+func (d setMinimumRequiredVersion) Transform(spec *cdi.Spec) error {
```
bart0sh, line 29: Is this an external API?

api/producer/writer.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	cdi "tags.cncf.io/container-device-interface/specs-go"
+)
+
+// A SpecWriter defines a structure for outputting CDI specifications.
+type SpecWriter struct {
```
bart0sh, line 29: Can this struct be non-public. It has only non-public fields.

pkg/cdi/cache.go
```diff
@@ -283,28 +284,35 @@ func (c *Cache) highestPrioritySpecDir() (string, int) {
 func (c *Cache) WriteSpec(raw *cdi.Spec, name string) error {
```
bart0sh, line 284: Should this API be moved to producer package?

bart0sh:
As we're adding new API package we should consider adding a documentation for it, similar to https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/doc.go

api/producer/api.go
elezar: I don't think so. `specs-go` defines the contents of the spec and not its representation on disk. Note that the `SpecFormat` type also has functionality for generating the contents which should not be in `specs-go`.

pkg/cdi/cache.go
elezar: Yes, `WriteSpec` is a producer function, but we can't move this to the `producer` package yet. Once we have the relevant validation logic in place (e.g. #247) we should be able to mark `WriteSpec` deprecated and / or remove it though.

api/producer/go.sum
```diff
@@ -0,0 +1,21 @@
...
+golang.org/x/mod v0.19.0/go.mod h1:hTbmBsO62+eylJbnUtE2MGJUyE7QWk4xUqPFrRgJ+7c=
+golang.org/x/sys v0.1.0 h1:kunALQeHf1/185U1i0GOB/fy1IPRDDpuoOOqRReG57U=
+golang.org/x/sys v0.1.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v2 v2.4.0 h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
+gopkg.in/yaml.v2 v2.4.0/go.mod h1:RDklbk79AGWmwhnvt/jBztapEOGDOx6ZbXqjP6csGnQ=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
+gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
```
bart0sh, line 19: I remember asking about this, but forgot the answer. Is there any reason to not use yaml.v3 for us?

api/producer/go.mod
```diff
@@ -0,0 +1,20 @@
...
+
+require (
+	github.com/stretchr/testify v1.7.0
+	golang.org/x/sys v0.1.0
+	sigs.k8s.io/yaml v1.3.0
+	tags.cncf.io/container-device-interface/specs-go v0.8.0
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
bart0sh, line 13: Is go-spew brought by testify?

api/producer/go.sum
elezar: I don't think so. I suppose we didn't want to introduce a dependency into our downstream clients that was not needed.

api/producer/writer.go
elezar: The issue is that we return this type from our `NewSpecWriter` constructure. We could return an interface instead, but I don't think that that's a requirement.

api/producer/set-minimum-version.go
elezar: It's not required to be public here. In the NVIDIA Container Toolkit implementation, we have the following interface:
```
type Transformer interface {
    Transform(*cdi.Spec) error
}
```
which allows for composition. 

I reused some of the implementation details, and didn't change the function visibility. 

Note that since the type itself is private, defining the function as public should not affect the API at all.

api/producer/go.mod
elezar: Yes, I believe so.

api/producer/writer.go
bart0sh: even if we return non-public struct it's still possible to call public methods on it IIRC.

pkg/cdi/cache.go
bart0sh: makes sense, thanks for the explanations!

api/producer/api.go
bart0sh: Fair enough, thanks!

api/producer/set-minimum-version.go
bart0sh: Yes, but it creates confusion. At least I was confused. Can we make it non-public?

pkg/cdi/spec.go
elezar: Updated.

api/producer/set-minimum-version.go
elezar: I will update it.

api/producer/writer.go
bart0sh: OK, makes sense.

pkg/cdi/cache.go
```diff
@@ -283,28 +284,35 @@ func (c *Cache) highestPrioritySpecDir() (string, int) {
...
 		return errors.New("no Spec directories to write to")
 	}
 
-	path = filepath.Join(specDir, name)
-	if ext := filepath.Ext(path); ext != ".json" && ext != ".yaml" {
-		path += defaultSpecExt
-	}
+	path := filepath.Join(specDir, name)
 
+	// We call newSpec to perform the required validation.
```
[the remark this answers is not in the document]
elezar: As discussed in the call, the point of the `error` being retunred from the constructor is largely to provide feedback on an invalid combination of options. This is also common in other projects such as k8s.

klihub:
One of the tests are failing now (TestCacheWriteSpec/multiple_spec_files/data,_some_invalid).

elezar:
> One of the tests is failing now (TestCacheWriteSpec/multiple_spec_files/data,_some_invalid).

Yes. I have updated the implementation. The `DefaultValidator` was not checking for non-null `Devices`.

Issue #240, opened 2024-11-11
Sort mounts with stable sort


Epsilon314:
We got an unexpected change of the order of mounts after introducing cdi (indirectly through updating Nvidia container toolkit), which in specific cases lead to failed container creation. The change of mounts order is related to the way sortMounts works, as referred below:
https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/container-edits.go#L358
```go
// sortMounts sorts the mounts in the given OCI Spec.
func sortMounts(specgen *ocigen.Generator) {
	mounts := specgen.Mounts()
	specgen.ClearMounts()
	sort.Sort(orderedMounts(mounts))
	specgen.Config.Mounts = mounts
}

// orderedMounts defines how to sort an OCI Spec Mount slice.
// This is the almost the same implementation sa used by CRI-O and Docker,
// with a minor tweak for stable sorting order (easier to test):
//
//	https://github.com/moby/moby/blob/17.05.x/daemon/volumes.go#L26
type orderedMounts []oci.Mount

// Len returns the number of mounts. Used in sorting.
func (m orderedMounts) Len() int {
	return len(m)
}

// Less returns true if the number of parts (a/b/c would be 3 parts) in the
// mount indexed by parameter 1 is less than that of the mount indexed by
// parameter 2. Used in sorting.
func (m orderedMounts) Less(i, j int) bool {
	ip, jp := m.parts(i), m.parts(j)
	if ip < jp {
		return true
	}
	if jp < ip {
		return false
	}
	return m[i].Destination < m[j].Destination
}
``` 

1. Go1.19 changes sort.Sort from stable to unstable, and we can use sort.Stable instead  (w.r.t https://github.com/containerd/containerd/blob/main/internal/cri/opts/spec_linux_opts.go#L68);
2. We can change orderedMounts.Less method to only compare parts, so that the mount order submitted by container spec yaml will be kept, in case it matters.

elezar:
@Epsilon314 thanks for calling this out. Do you have example mounts so that we can include those cases in our tests while addressing this?

Epsilon314:
@elezar  We encounter this issue when creating a docker in docker container:
1. Mount to container path `/run/containerd` and `/proc/cpuinfo`
[...]

PR #241, opened 2024-11-12
sortMounts stable and remove alphabetical order

Epsilon314:
PR for  #240

### DIFF
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -355,7 +355,7 @@ func ensureOCIHooks(spec *oci.Spec) {
 func sortMounts(specgen *ocigen.Generator) {
 	mounts := specgen.Mounts()
 	specgen.ClearMounts()
-	sort.Sort(orderedMounts(mounts))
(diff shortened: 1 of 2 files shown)


elezar (commented):
Thanks @Epsilon314

What is the expected behaviour when the mount being injected has:
1. the same number of parts
2. a greater number of parts

The behaviour for 1 seems undefined and was the motivation for the alphabetical sorting in the first place.

Epsilon314:
@elezar For mounts with same number of parts, its original order will be kept (stable sort). Since containerd also uses stable sort, users can therefore arrange the order their mounts (of same number of parts ) in container spec if they do need a specific order.

elezar:
Ok, so if we add mounts these would always be added after the existing mounts. Could you add tests for those cases?

Epsilon314:
> Ok, so if we add mounts these would always be added after the existing mounts. Could you add tests for those cases?

@elezar OK

klihub (approved):

bart0sh (approved):

PR #242, opened 2024-11-25
Use existing go version in go mod tidy an verify

elezar:
This change ensures that running make mod-tidy doesn't update the go version specified in the mod files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ $(BINARIES): bin/%:
 mod-tidy:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Tidying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod tidy \
+	        cd $$(dirname $$mod) && go mod tidy -go=$$(grep -E "^go\s+[1-9]\.[0-9]+$$" go.mod | sed 's/^go //') \
             ) || exit 1; \
 	done
 


Makefile
```diff
@@ -74,14 +74,14 @@ $(BINARIES): bin/%:
 mod-tidy:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Tidying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod tidy \
+	        cd $$(dirname $$mod) && go mod tidy -go=$$(grep -E "go\s+\d\.\d+" go.mod | sed 's/go //') \
```
klihub, line 77: nit: maybe it's just me but my gut feeling would like 'go ' pattern to be forced to only match at the beginning of the line  with "^go...".

Makefile
```diff
@@ -74,14 +74,14 @@ $(BINARIES): bin/%:
...
             ) || exit 1; \
 	done
 
 mod-verify:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Verifying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod verify | sed 's/^/  /g' \
+	        cd $$(dirname $$mod) && go mod verify -go=$$(grep -E "go\s+\d\.\d+" go.mod | sed 's/go //') | sed 's/^/  /g' \
```
klihub, line 84: nit: ditto here.

Makefile
```diff
@@ -74,14 +74,14 @@ $(BINARIES): bin/%:
 mod-tidy:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Tidying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod tidy \
+	        cd $$(dirname $$mod) && go mod tidy -go=$$(grep -E "^go\s+\d\.\d+$$" go.mod | sed 's/^go //') \
```
bart0sh, line 77: `grep -E "^go\s+\d\.\d+$$" go.mod` doesn't output anything in my env:
```
$ grep -E "^go\s+\d\.\d+$" go.mod || echo nope
nope
```

I think `grep` doesn't understand `\d `pattern, does it?

elezar:
```
container-device-interface git:(producer-api) ✗ grep -E "^go\s+\d\.\d+$" go.mod | sed 's/^go //'
1.20
```

elezar:
```➜  container-device-interface git:(fix-go-mod-tidy) ✗ docker run --rm -ti -v $(pwd):/work -w /work ubuntu grep -E "^go\s+[1-9]\.[0-9]+$" go.mod
go 1.20
➜  container-device-interface git:(fix-go-mod-tidy) ✗ grep -E "^go\s+[1-9]\.[0-9]+$" go.mod
go 1.20```

File: Makefile [hunk 4]
```diff
@@ -74,14 +74,14 @@ $(BINARIES): bin/%:
 mod-tidy:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Tidying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod tidy \
+	        cd $$(dirname $$mod) && go mod tidy -go=$$(grep -E "^go\s+[1-9]\.[0-9]+$$" go.mod | sed 's/^go //') \
```
bart0sh, line 77: I'm still not sure I understand the intent of the change.
According to the `go mod tidy` help output
```
The -go flag causes tidy to update the 'go' directive in the go.mod
file to the given version
```

However, updating the directive to the version from the same file does nothing, doesn't it?

Makefile
```diff
@@ -74,14 +74,14 @@ $(BINARIES): bin/%:
...
             ) || exit 1; \
 	done
 
 mod-verify:
 	$(Q)for mod in $$(find . -name go.mod); do \
 	    echo "Verifying $$mod..."; ( \
-	        cd $$(dirname $$mod) && go mod verify | sed 's/^/  /g' \
+	        cd $$(dirname $$mod) && go mod verify -go=$$(grep -E "^go\s+[1-9]\.[0-9]+$$" go.mod | sed 's/^go //') | sed 's/^/  /g' \
```
bart0sh, line 84: `go mod verify` doesn't have `-go` command line option according to [its help page](https://go.dev/ref/mod#go-mod-verify).

elezar: You're right! Updated.

elezar:
Actually, I think I missed something here. There reason my `go` version was being updated was that I had inadvertently created a go.mod file in `api/producer` with `go 1.23.2` (running `go mod init`). This is then propagated to modules that pull this in.

I'm going to close this PR as unneeded since setting `go 1.20` in that module addresses the issue.

Sorry for the back and forth.

PR #248, opened 2025-02-04
Create schema submodule

elezar:
The github.com/xeipuuv/gojsonschema that is imported for schema validation can be problematic.

This change creates a go submodule at `schema` to remove this dependency from the main module.

Note that this also moves from a function-based schema validation to using an interface so that the schema-based validators are pluggable into the producer as implemented in #233


klihub (dismissed):

pkg/cdi/cache_test.go
```diff
@@ -225,6 +225,9 @@ devices:
 }
 
 func TestRefreshCache(t *testing.T) {
+	if runtime.GOOS == "darwin" {
+		t.Skip("skipping on darwin")
+	}
```
bart0sh, line 230: Can you explain why we're skipping this test on Darwin?

bart0sh:
@elezar schema unit tests fail to build it seems:
```
Running tool: /usr/local/go/bin/go test -timeout 300s -coverprofile=/tmp/vscode-goYWugaU/go-code-cover tags.cncf.io/container-device-interface/schema

# tags.cncf.io/container-device-interface/schema_test [tags.cncf.io/container-device-interface/schema.test]
/home/ed/git/container-device-interface/schema/schema_test.go:301:22: cannot use bytes.NewReader(data) (value of type *bytes.Reader) as *"tags.cncf.io/container-device-interface/specs-go".Spec value in argument to scm.Validate
/home/ed/git/container-device-interface/schema/schema_test.go:303:16: undefined: schema.Validate
/home/ed/git/container-device-interface/schema/schema_test.go:318:22: cannot use r (variable of type io.Reader) as *"tags.cncf.io/container-device-interface/specs-go".Spec value in argument to scm.Validate
/home/ed/git/container-device-interface/schema/schema_test.go:320:16: undefined: schema.Validate
/home/ed/git/container-device-interface/schema/schema_test.go:326:22: cannot use bytes.NewReader(buf.Bytes()) (value of type *bytes.Reader) as *"tags.cncf.io/container-device-interface/specs-go".Spec value in argument to scm.Validate
/home/ed/git/container-device-interface/schema/schema_test.go:328:16: undefined: schema.Validate
FAIL	tags.cncf.io/container-device-interface/schema [build failed]
FAIL
```

pkg/cdi/cache_test.go
elezar: I didn't look into it too deeply, but my assumption was that the ordering is different there causing the refreshed state to not be the same as encoded in the tests.

I can update the message to make this clearer if you prefer. (to be clear this is actually part of #249)

bart0sh:
@elezar There is something wrong with the package dependencies, I guess. `go list` doesn't show all of them. This could be a reason why go test ./... doesn't run all tests.
```
$ go list ./...
tags.cncf.io/container-device-interface/internal/validation
tags.cncf.io/container-device-interface/internal/validation/k8s
tags.cncf.io/container-device-interface/pkg/cdi
tags.cncf.io/container-device-interface/pkg/parser
```

bart0sh:
why schema_test.go uses it's own package `schema_test` ?

Makefile
```diff
@@ -103,7 +103,11 @@ clean-schema:
 
 # tests for go packages
 test-gopkgs:
-	$(Q)$(GO_TEST) ./...
+	$(Q)for mod in $$(find . -name go.mod); do \
+	    echo "Testing $$(dirname $$mod)..."; ( \
+	        cd $$(dirname $$mod) && $(GO_TEST) ./... \
+            ) || exit 1; \
```
bart0sh, line 109: Would it exit on the first test failure? If so, could it be modified to run all tests even if one fails?

elezar: I have updated this to run all tests and exit on failures afterwards.

klihub:
> > why schema_test.go uses its own package `schema_test` ?
> 
> I didn't change the test package as part of this change. Maybe the original intent was to test against public interfaces. @klihub do you recall if there was a reason for this?

Yes, we do blackbox testing using the public interfaces only. It's normal go practice to put the tests in ${package}_test in that case.

bart0sh:
> Yes, we do blackbox testing using the public interfaces only. It's normal go practice to put the tests in ${package}_test in that case.

Is this package somehow different from others in this project?

$ find . -name *_test.go |xargs grep package
./pkg/parser/parser_test.go:package parser
./pkg/cdi/default-cache_test.go:package cdi
./pkg/cdi/regressions_test.go:package cdi
./pkg/cdi/annotations_test.go:package cdi
./pkg/cdi/container-edits_test.go:package cdi
./pkg/cdi/spec-dirs_test.go:package cdi
./pkg/cdi/device_test.go:package cdi
./pkg/cdi/cache_test.go:package cdi
./pkg/cdi/spec_test.go:package cdi
**./schema/schema_test.go:package schema_test**

pkg/cdi/cache.go
```diff
@@ -529,17 +530,24 @@ func (w *watch) watch(fsw *fsnotify.Watcher, m *sync.Mutex, refresh func() error
 	if watch == nil {
 		return
 	}
+
+	eventMask := fsnotify.Rename | fsnotify.Remove | fsnotify.Write
+	// On macOS, we also need to watch for Create events.
```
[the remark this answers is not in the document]
klihub: Ah, ok. So its rename(2) which behaves differently wrt. fsnotify on macOS.

klihub:
> > Yes, we do blackbox testing using the public interfaces only. It's normal go practice to put the tests in ${package}_test in that case.
> 
> Is this package somehow different from others in this project?
> 
> $ find . -name *_test.go |xargs grep package ./pkg/parser/parser_test.go:package parser ./pkg/cdi/default-cache_test.go:package cdi ./pkg/cdi/regressions_test.go:package cdi ./pkg/cdi/annotations_test.go:package cdi ./pkg/cdi/container-edits_test.go:package cdi ./pkg/cdi/spec-dirs_test.go:package cdi ./pkg/cdi/device_test.go:package cdi ./pkg/cdi/cache_test.go:package cdi ./pkg/cdi/spec_test.go:package cdi **./schema/schema_test.go:package schema_test**

I think many of the tests in pkg/cdi make minimal use of some unexported functions (for instance newCache, newSpec, validation functions, those sort of things) and that's why they were made part of the package they test. The schema package is mostly a pile of boilerplate with the specific purpose to hide isolate/hide the xeipuuv/gojsonschema interfaces and transform them to something behind which we could plug in another implementation if/once we found one more adequate one. So there it was an obvious choice to go for blackbox testing.

bart0sh:
@klihub Thank you for the explanation! Now that makes sense to me. `parser` tests can also use its own package it seems.

elezar:
> @klihub Thank you for the explanation! Now that makes sense to me. `parser` tests can also use its own package it seems.

Sure, but that is out of scope for this PR. I'm happy to address that as a follow-up.

bart0sh (approved):

klihub (approved):

PR #249, opened 2025-02-04
Fix cache refresh on darwin

elezar:
These changes fix running the unit tests on MacOS and make some minor golangci-lint changes.

The core of the problem is that file creation using `os.Rename` on `darwin` generates `CREATE` events and not `WRITE` events. This change also responds to CREATE events when updating the cache.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -22,6 +22,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -529,6 +530,13 @@ func (w *watch) watch(fsw *fsnotify.Watcher, m *sync.Mutex, refresh func() error
 	if watch == nil {
 		return
 	}
+
+	eventMask := fsnotify.Rename | fsnotify.Remove | fsnotify.Write
+	// On macOS, we also need to watch for Create events.
+	if runtime.GOOS == "darwin" {
+		eventMask |= fsnotify.Create
+	}
+
 	for {
 		select {
 		case event, ok := <-watch.Events:
@@ -536,10 +544,10 @@ func (w *watch) watch(fsw *fsnotify.Watcher, m *sync.Mutex, refresh func() error
 				return
 			}
 
-			if (event.Op & (fsnotify.Rename | fsnotify.Remove | fsnotify.Write)) == 0 {
+			if (event.Op & eventMask) == 0 {
 				continue
 			}
-			if event.Op == fsnotify.Write {
+			if event.Op == fsnotify.Write || event.Op == fsnotify.Create {
 				if ext := filepath.Ext(event.Name); ext != ".json" && ext != ".yaml" {
 					continue
 				}
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -11,6 +11,8 @@ permissions:
   contents: read
   # Optional: allow read access to pull request. Use with `only-new-issues` option.
   # pull-requests: read
+  # Optional: allow write access to checks to allow the action to annotate code in the PR.
+  checks: write
 
 jobs:
   golangci:
--- a/pkg/cdi/cache_test_darwin.go
+++ b/pkg/cdi/cache_test_darwin.go
@@ -1,5 +1,5 @@
-//go:build !windows && !darwin
-// +build !windows,!darwin
+//go:build darwin
+// +build darwin
 
 /*
    Copyright © 2021 The CDI Authors
@@ -22,5 +22,5 @@ package cdi
 import "syscall"
 
 func osSync() {
-	syscall.Sync()
+	_ = syscall.Sync()
 }
--- a/pkg/cdi/container-edits_test.go
+++ b/pkg/cdi/container-edits_test.go
@@ -17,6 +17,7 @@
 package cdi
 
 import (
+	"runtime"
 	"testing"
 
 	oci "github.com/opencontainers/runtime-spec/specs-go"
@@ -303,6 +304,13 @@ func TestValidateContainerEdits(t *testing.T) {
 }
 
 func TestApplyContainerEdits(t *testing.T) {
+	nullDeviceMajor := int64(1)
+	nullDeviceMinor := int64(3)
+	if runtime.GOOS == "darwin" {
+		nullDeviceMajor = 3
+		nullDeviceMinor = 2
+	}
+
 	type testCase struct {
 		name   string
 		spec   *oci.Spec
@@ -346,8 +354,8 @@ func TestApplyContainerEdits(t *testing.T) {
 						{
 							Path:  "/dev/null",
 							Type:  "c",
-							Major: 1,
-							Minor: 3,
+							Major: nullDeviceMajor,
+							Minor: nullDeviceMinor,
 						},
 					},
 					Resources: &oci.LinuxResources{
@@ -355,8 +363,8 @@ func TestApplyContainerEdits(t *testing.T) {
 							{
 								Allow:  true,
 								Type:   "c",
-								Major:  int64ptr(1),
-								Minor:  int64ptr(3),
+								Major:  &nullDeviceMajor,
+								Minor:  &nullDeviceMinor,
 								Access: "rwm",
 							},
 						},
@@ -389,8 +397,8 @@ func TestApplyContainerEdits(t *testing.T) {
 						{
 							Path:  "/dev/null",
 							Type:  "c",
-							Major: 1,
-							Minor: 3,
+							Major: nullDeviceMajor,
+							Minor: nullDeviceMinor,
 						},
 					},
 					Resources: &oci.LinuxResources{
@@ -398,8 +406,8 @@ func TestApplyContainerEdits(t *testing.T) {
 							{
 								Allow:  true,
 								Type:   "c",
-								Major:  int64ptr(1),
-								Minor:  int64ptr(3),
+								Major:  &nullDeviceMajor,
+								Minor:  &nullDeviceMinor,
 								Access: "rwm",
 							},
 						},
--- a/pkg/cdi/spec.go
+++ b/pkg/cdi/spec.go
@@ -259,7 +259,7 @@ func SetSpecValidator(fn func(*cdi.Spec) error) {
 	specValidator = fn
 }
 
-// validateSpec validates the Spec using the extneral validator.
+// validateSpec validates the Spec using the external validator.
 func validateSpec(raw *cdi.Spec) error {
 	validatorLock.RLock()
 	defer validatorLock.RUnlock()


klihub (approved):

klihub (approved):
LGTM.

nit: I think a `git rebase 081e75b3d69ed0911640e92d9b0af41181934a25 --onto 327c7f88092fbe9e378b8b3b5ef87a6633bd4f48` before pushing would have made it more sha1-obvious that #248 and #249 indeed share the bottom-most commit.

pkg/cdi/default-cache_test.go
```diff
@@ -129,6 +130,9 @@ devices:
 }
 
 func TestDefaultCacheRefresh(t *testing.T) {
+	if runtime.GOOS == "darwin" {
```
elezar, line 133: This fails with:
```
```
=== RUN   TestDefaultCacheRefresh/empty_cache,_add_one_Spec
    default-cache_test.go:292:
                Error Trace:    default-cache_test.go:292
                Error:          Not equal:
                                expected: []string{"vendor1.com/device=dev1"}
                                actual  : []string(nil)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,2 @@
                                -([]string) (len=1) {
                                - (string) (len=23) "vendor1.com/device=dev1"
                                -}
                                +([]string) <nil>

                Test:           TestDefaultCacheRefresh/empty_cache,_add_one_Spec
```

File: pkg/cdi/cache_test.go [hunk 2]
```diff
@@ -225,6 +226,9 @@ devices:
 }
 
 func TestRefreshCache(t *testing.T) {
+	if runtime.GOOS == "darwin" {
```
elezar, line 229: fails with:
```
=== RUN   TestRefreshCache/empty_cache,_add_one_Spec
    cache_test.go:581:
                Error Trace:    cache_test.go:581
                Error:          Not equal:
                                expected: []string{"vendor1.com/device=dev1"}
                                actual  : []string(nil)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,2 @@
                                -([]string) (len=1) {
                                - (string) (len=23) "vendor1.com/device=dev1"
                                -}
                                +([]string) <nil>

                Test:           TestRefreshCache/empty_cache,_add_one_Spec
=== RUN   TestRefreshCache/one_Spec,_add_another,_no_shadowing,_no_conflicts
    cache_test.go:581:
                Error Trace:    cache_test.go:581
                Error:          Not equal:
                                expected: []string{"vendor1.com/device=dev1", "vendor1.com/device=dev2"}
                                actual  : []string{"vendor1.com/device=dev1"}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,3 @@
                                -([]string) (len=2) {
                                - (string) (len=23) "vendor1.com/device=dev1",
                                - (string) (len=23) "vendor1.com/device=dev2"
                                +([]string) (len=1) {
                                + (string) (len=23) "vendor1.com/device=dev1"
                                 }
                Test:           TestRefreshCache/one_Spec,_add_another,_no_shadowing,_no_conflicts
=== RUN   TestRefreshCache/two_Specs,_remove_one
=== RUN   TestRefreshCache/one_Spec,_add_another,_shadowing
    cache_test.go:587:
                Error Trace:    cache_test.go:587
                Error:          Not equal:
                                expected: 0
                                actual  : 1
                Test:           TestRefreshCache/one_Spec,_add_another,_shadowing
=== RUN   TestRefreshCache/one_Spec,_add_another,_conflicts
    cache_test.go:581:
                Error Trace:    cache_test.go:581
                Error:          Not equal:
                                expected: []string{"vendor1.com/device=dev2", "vendor1.com/device=dev3"}
                                actual  : []string{"vendor1.com/device=dev1", "vendor1.com/device=dev2"}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,4 @@
                                 ([]string) (len=2) {
                                - (string) (len=23) "vendor1.com/device=dev2",
                                - (string) (len=23) "vendor1.com/device=dev3"
                                + (string) (len=23) "vendor1.com/device=dev1",
                                + (string) (len=23) "vendor1.com/device=dev2"
                                 }
                Test:           TestRefreshCache/one_Spec,_add_another,_conflicts
--- FAIL: TestRefreshCache (0.70s)
    --- FAIL: TestRefreshCache/empty_cache,_add_one_Spec (0.11s)
    --- FAIL: TestRefreshCache/one_Spec,_add_another,_no_shadowing,_no_conflicts (0.13s)
    --- PASS: TestRefreshCache/two_Specs,_remove_one (0.15s)
    --- FAIL: TestRefreshCache/one_Spec,_add_another,_shadowing (0.15s)
    --- FAIL: TestRefreshCache/one_Spec,_add_another,_conflicts (0.16s)
```

pkg/cdi/cache.go
```diff
@@ -529,17 +530,24 @@ func (w *watch) watch(fsw *fsnotify.Watcher, m *sync.Mutex, refresh func() error
 	if watch == nil {
 		return
 	}
+
+	eventMask := fsnotify.Rename | fsnotify.Remove | fsnotify.Write
```
elezar, line 534: @bart0sh after your probing I spent some more time investigating this, and noted that the events that are generated on `darwin` are `CREATE` events and not `WRITE` events.

I thought it safer to not modify existing behaviour on non-darwin platforms.

bart0sh: Thanks for investigation. I'm wondering why CREATE was not considered on Linux. I believe it's also generated when file is created. Do we intentionally skip it?

elezar: @klihub since you were the one who implemented this, maybe you could comment.

Note that `fsnotify` uses two different sources for the events on Linux and MacOS systems. It may also be that the `os.Rename` calls that are use in the tests trigger different events.

klihub: > Thanks for investigation. I'm wondering why CREATE was not considered on Linux. I believe it's also generated when file is created. Do we intentionally skip it?

On Linux,
  - when you create a non-empty file you always get an fsnotify.Write event (in addition to the create)
  - when you rename a file, you always get at least an fsnotify.Rename event (plus a create if the renamed file is in the same directory)
Therefore, it should not be necessary to trigger a rescan for a fsnotify.Create.

pkg/cdi/cache.go
```diff
@@ -529,17 +530,24 @@ func (w *watch) watch(fsw *fsnotify.Watcher, m *sync.Mutex, refresh func() error
...
+	// On macOS, we also need to watch for Create events.
+	if runtime.GOOS == "darwin" {
+		eventMask |= fsnotify.Create
```
klihub, line 537: nit: if Darwin behaves the same way as Linux, IOW a rename to the same directory generates fsnotify.Rename+fsnotify.Create and a rename to a different directory generates an fsnotify.Rename, then I think it should not be necessary to watch for fsnotify.Create.

klihub (approved):
LGTM.

@elezar I have a small questio-nit, but even with that considered this looks good to me.

pkg/cdi/cache.go
bart0sh: Interesting, does it mean that we don't need this `if` statement at all?

bart0sh: @klihub events are the same, but their order is different. On Linux it generates fsnotify.Rename+fsnotify.Create, on Darwin: fsnotify.Create+fsnotify.Rename.

elezar: The specific issue in the tests is that we:
1. Start a watch on a directory
2. Create a file `foo.yaml.tmp` in that folder
3. Write to `foo.yaml.tmp`.
4. Rename that file to `foo.yaml`

This was *not* triggering the refresh in the auto refresh case.

elezar: Let's assume the following example (adapted from the [test code](https://github.com/cncf-tags/container-device-interface/blob/4149a8a524dfe96686d97789833cc78f61eb6631/pkg/cdi/spec-dirs_test.go#L229-L247)):
```
package main

import (
	"log"
	"os"
	"path/filepath"
)

func main() {
	dir := "/tmp/fsnotify"
	filename := "test.txt"

	data := "hello world"

	file := filepath.Join(dir, filename)
	tmp := file + ".tmp"
	if err := os.WriteFile(tmp, []byte(data), 0644); err != nil {
		log.Fatalf("failed to write file %q: %w", tmp, err)
	}
	if err := os.Rename(tmp, file); err != nil {
		log.Fatalf("failed to rename %q to %q: %w", tmp, file, err)
	}
}
```

Using the example from: https://github.com/fsnotify/fsnotify/tree/main/cmd/fsnotify I run:
```
./fsnotify watch /tmp/fsnotify
```

When I run the example above on `darwin`, I see:
```
./fsnotify watch /tmp/fsnotify
22:14:04.7182 ready; press ^C to exit
22:17:11.7594   1 CREATE        "/tmp/fsnotify/test.txt"
```

Repeating in a linux container, I see:
```
root@7ff8b7d75a3c:/work# ./fsnotify watch /tmp/fsnotify
21:22:25.3845 ready; press ^C to exit
21:22:52.6097   1 CREATE        "/tmp/fsnotify/test.txt.tmp"
21:22:52.6097   2 WRITE         "/tmp/fsnotify/test.txt.tmp"
21:22:52.6098   3 RENAME        "/tmp/fsnotify/test.txt.tmp"
21:22:52.6098   4 CREATE        "/tmp/fsnotify/test.txt" ← "/tmp/fsnotify/test.txt.tmp"
```

Note that when looking at the event loop, we see:
```
			if !ok {
				return
			}

			if (event.Op & eventMask) == 0 {
				continue
			}
			if event.Op == fsnotify.Write || event.Op == fsnotify.Create {
				if ext := filepath.Ext(event.Name); ext != ".json" && ext != ".yaml" {
					continue
				}
			}

			m.Lock()
			if event.Op == fsnotify.Remove && w.tracked[event.Name] {
				w.update(dirErrors, event.Name)
			} else {
				w.update(dirErrors)
			}
			_ = refresh()
			m.Unlock()
```
meaning that it's the `RENAME` that is triggering the `w.update(dirErrors)` *unconditionally* in this case.

Note that the `CREATE` added in the darwin case is what is needed to trigger the refresh there.

elezar: To dig a bit further:

I ran the following commands (where `/tmp/fsnotify/test.txt` already exists):
```
cd /tmp/fsnotify
rm test.txt
touch touch.txt
echo "echo" > echo.txt
mv touch.txt mv.txt
```
### Darwin:
```
22:30:12.6052 ready; press ^C to exit
22:30:29.9514   1 REMOVE        "/tmp/fsnotify/test.txt"
22:30:43.7353   2 CREATE        "/tmp/fsnotify/touch.txt"
22:30:53.8444   3 CREATE        "/tmp/fsnotify/echo.txt"
22:31:09.0065   4 CREATE        "/tmp/fsnotify/mv.txt"
22:31:09.0066   5 RENAME        "/tmp/fsnotify/touch.txt"
```
### Linux:
```
21:34:29.6950 ready; press ^C to exit
21:34:36.3417   1 REMOVE        "/tmp/fsnotify/test.txt"
21:34:46.2412   2 CREATE        "/tmp/fsnotify/touch.txt"
21:34:46.2413   3 CHMOD         "/tmp/fsnotify/touch.txt"
21:34:51.0527   4 CREATE        "/tmp/fsnotify/echo.txt"
21:34:51.0529   5 WRITE         "/tmp/fsnotify/echo.txt"
21:34:54.8228   6 RENAME        "/tmp/fsnotify/touch.txt"
21:34:54.8229   7 CREATE        "/tmp/fsnotify/mv.txt" ← "/tmp/fsnotify/touch.txt"
```

Note that:
* There are *no* `WRITE` events genereated on `darwin`
* As @bart0sh points out, the ordering of `RENAME` and `CREATE` on a move is reversed.
* An `os.Rename()` call is not the same as a `mv` on `darwin`.

From this, it seems as if we could rely on only `CREATE` and `REMOVE` on call platforms, with the caveat that we may be calling refresh on incomplete data due to the `CREATE` being triggered before a `WRITE` to the same file. This is an issue for a `WRITE` in any case since this does not guarantee that the file is complete.

bart0sh: [unix.Renameat2](https://pkg.go.dev/golang.org/x/sys/unix#Renameat2) generates the same events as [os.Rename](https://pkg.go.dev/os#Rename):  fsnotify.Rename+fsnotify.Create.

bart0sh: @elezar 
> There are no WRITE events genereated on darwin

Yep, but it's generated if the file already exists. We should probably consider testing it with the cache API calls. It could also make sense to modify it to use `os.Create` before `ioutil.WriteFile` to force generation of the `WRITE` event if we want the same code to work on Darwin and Linux.

BTW, since then CDI supports Darwin? I was under impression that only Linux is supported. We should explicitly mention this in the docs, I believe.

klihub: @elezar I am fine with the current state where on Darwin we also watch for the create event, if that helps your daily workflow which I gather is on MacOS/Darwin. IMO, this is not necessary on Linux and not helpful either (it would only generate an extra event if someone created an empty file without ever writing any content to it).

About the rest of the discussion related to trying to write/create files differently to generate a different set, or order, of inotify/fsnotify events... I think that is a bit futile. As @elezar pointed out, we don't control how users write Spec files and neither should we try. Spec reading/parsing is supposed to be fault tolerant for multiple reasons, the uncontrollability and thus unpredictability of the set and order of events generated being one of them. Therefore, as long as we make sure that we never fail to act on an event eventually, and we don't delay acting on events too long (this is the reason why we don't have an timer-based event filtering/deduplication) we should be fine.

klihub: > @klihub events are the same, but their order is different. On Linux it generates fsnotify.Rename+fsnotify.Create, on Darwin: fsnotify.Create+fsnotify.Rename.

On linux, both events are generated by the time the file has been renamed. On Darwin, I'm not sure, although one would hope that a rename(2) and related variants would be atomic (IIRC, required for POSIX-compliance) which should mean that both are generated after the fact, otherwise it is not really atomic if an intermediate state is externally observable. Then again, since the tests did fail on Darwin and I assume the failing tests check the right thing to decide success/failure, some of this mut be false on Darwin... but then based on @elezar's experience and fix it looks like adding an extra watch on Darwin for fsnotify.Create fixes things there... so I think this should be now fine there, too.

pkg/cdi/cache.go
jgehrcke: >  I spent some more time investigating this, and noted that the events that are generated on darwin are CREATE events and not WRITE events

Nice.

>  two different sources for the events on Linux and MacOS systems. 

Trivia. macOS itself provides two different mechanisms: FSEvents, kqueue. fsnotify seems to use kqueue.

Linux has its own mechanism, called [inotify](https://linux.die.net/man/7/inotify). [Here](https://news.ycombinator.com/item?id=8669235) is a bikeshed thread comparing both interfaces.

I like the docs of https://pythonhosted.org/watchdog/installation.html#supported-platforms-and-caveats.

Interesting: https://watchexec.github.io/docs/macos-fsevents.html

PR #252, opened 2025-02-17
Bump spec version to v0.9.0

elezar:


specs-go/version.go
```diff
@@ -138,9 +140,16 @@ func (r requiredVersionMap) requiredVersion(spec *Spec) version {
 	return minVersion
 }
 
+// requiresV090 returns true if the spec uses v0.9.0 features.
+// Since the v0.9.0 spec bump was due to moving the minimum version checks to
+// the spec package, there are no explicit spec changes.
+func requiresV090(_ *Spec) bool {
+	return false
+}
+
```
klihub, line 149: This is just out of curiosity: is there any difference between `validSpecVersions[$VERSION]` being `nil` and it being a function which always returns `false` ?

elezar: There is no difference functionally. We could remove the `requiresV080` and `requiresV090` functions entirely.

klihub: Ok. Just wanted to make sure I understand it correctly. From my perspective, there is no need to remove those functions to get this merged. There is one benefit to having those functions there as opposed to omitting them: it spells out that this bit of functionality has been considered/taken care of, but is just happens that there are no new additions to the Spec in those versions.

bart0sh:
@elezar I'd propose to close this PR in favour of https://github.com/cncf-tags/container-device-interface/pull/257 as we agreed to release current main branch as 1.0.0 recently.

bart0sh:
Closing as #257 is merged.

PR #254, opened 2025-02-22
fix: don't crash in cache refresh/update with nil fsnotify watcher.

### LINKED ISSUE #253: Cache SIGSEGV panic on refresh if fsnotify NewWatch() fails, for instance due to too many open file descriptors.
See https://github.com/moby/buildkit/issues/5767 for details.

klihub:
Don't crash if fsnotify.Watcher creation fails, for instance due to being out of file descriptors. See https://github.com/moby/buildkit/issues/5767 for an example.

Fixes #253.


pkg/cdi/cache.go
```diff
@@ -579,6 +579,14 @@ func (w *watch) update(dirErrors map[string]error, removed ...string) bool {
 		update bool
 	)
 
+	// If we failed to create an fsnotify.Watcher we have a nil watcher here
+	// (but with autoRefresh left on). One known case when this can happen is
+	// if we have too many open files. In that case we always return true and
+	// force a refresh.
+	if w.watcher == nil {
```
bart0sh, line 586: Is ti possible to check it earlier, e.g. when `w.watcher` [is about to be assigned](https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/cache.go#L501) to `nil`? Or we want to keep it nil in a hope that it will be changed at some point?

pkg/cdi/cache_test.go
```diff
@@ -812,6 +814,102 @@ devices:
 	}
 }
 
+func TestTooManyOpenFiles(t *testing.T) {
+	if runtime.GOOS != "linux" {
```
bart0sh, line 818: Would it make sense to move this test to the `cache_test_linux.go` ? Would it work on other *nixes or Darwin?

pkg/cdi/cache_test.go
```diff
@@ -812,6 +814,102 @@ devices:
...
+
+	if err := syscall.Setrlimit(syscall.RLIMIT_NOFILE, &em.limit); err != nil {
+		return err
+	}
+	for _, fd := range em.fds {
+		syscall.Close(fd)
+	}
+
+	return nil
+}
```
bart0sh, line 911: It would be good to have some description/comments for this code.

pkg/cdi/cache_test.go
```diff
@@ -812,6 +814,102 @@ devices:
...
+	_, err = syscall.Socket(syscall.AF_INET, syscall.SOCK_DGRAM, 0)
+	require.Equal(t, syscall.EMFILE, err)
+
+	cache := newCache(
+		WithAutoRefresh(true),
+	)
+	require.NotNil(t, cache)
+
+	// try to trigger original crash with a nil fsnotify.Watcher
+	_, _ = cache.InjectDevices(&oci.Spec{}, "vendor1.com/device=dev1")
```
bart0sh, line 838: nit:
```suggestion
	devs := []string{"vendor1.com/device=dev1"}
	devices, err := cache.InjectDevices(&oci.Spec{}, devs...)
	require.Error(t, err)
	require.EqualValues(t, devs, devices)
```

Would it make sense to test other `cache` APIs as well?

pkg/cdi/cache.go
elezar: If I recall correctly we guard setting the watchers in a Mutex. Does that help to ensure this does not happen?

elezar (approved):
Thanks @klihub.

I think the changes look good in principle.

@bart0sh feel free to merge once you're happy so that we can create a branch and cherry-pick this change and possibly the mount ordering one.

pkg/cdi/cache_test.go
klihub: It could be made to work on other un*xen as well. Would need an OS-specific implementation (and a renaming) of getProcStatusFdSize. But it probably does not make much sense at the moment so we might as well move all the related code to cache_test_linux.go.

pkg/cdi/cache.go
klihub: > Is it possible to check it earlier, e.g. when `w.watcher` [is about to be assigned](https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/cache.go#L501) to `nil`? Or we want to keep it nil in a hope that it will be changed at some point?

Yes, I wanted to leave the door open for a better than the current recovery once we can again create fds (and that would probably be retrying a setup here).

In this very PR, I wanted to roll a minimum-footprint fix that we can review and merge quickly, and be fairly confident that it is safe to cherry pick and tag a v0.8.1.

klihub: > If I recall correctly we guard setting the watchers in a Mutex. Does that help to ensure this does not happen?

This happens if we hit EMFILE while trying to create the watch.

pkg/cdi/cache_test.go
klihub: Or cache_linux_test.go...

pkg/cdi/cache_test.go
klihub: Updated with comments about how we exhaust file descriptors.

pkg/cdi/cache_test.go
klihub: That InjectDevices should and will fail her (we have an empty cache, and we also have our file descriptors exhausted) but it should do it gracefully instead for crashing with a SIGSEGV (which it does until the next/tipmost commit is applied).

pkg/cdi/cache_test.go
klihub: Moved to a linux-specific test file.

pkg/cdi/cache_test.go
klihub: Added a TestRecoveryAfterTooManyOpenFiles test case to verify that the cache becomes functional again if the process ever recovers from its inability to create file descriptors.

Issue #259, opened 2025-03-07
Release container-device-interface v1.0.0


bart0sh:
- [x] Create a PR titled `Bump version to v1.0.0` including the following changes:
    - [x] Change the following to the target version `v1.0.0`:
        - [x] the `CurrentVersion` in `specs-go/versions.go`,
        - [x] the `tags.cncf.io/container-device-interface` version in `schema/go.mod`,
        - [x] (*for specification changes only*) the `tags.cncf.io/container-device-interface/specs-go` version in go.mod,
        - [x] (*for specification changes only*) the `tags.cncf.io/container-device-interface/specs-go` version in `schema/go.mod`.
    - [x] Run `make mod-tidy` to update versions in `cmd/**/go.mod`.
    - [x] Run `make mod-verify` to ensure modules are up to date.
    - [x] (*for specification changes only*) Add a description to the specification changes in `SPEC.md`.
    - [x] (*for specification changes only*) Implement a `requiresV*` function for the target version in `specs-go/versions.go`.
- [x] Merge the PR on sufficient approval.
- [x] Create a `v1.0.0` tag.
- [x] (*for specification changes only*) Create a `specs-go/v1.0.0` tag. (for the first spec version `w` will be the same as `z`)
- [x] Create a GitHub release form the `v1.0.0` tag.
- [x] (*for specification changes only*) Create a GitHub release from the `specs-go/v1.0.0` tag.
- [x] Close the release issue.

elezar:
Thanks @bart0sh.

How do we feel about a script (e.g. `/hack/prepare-release.sh`) that uses the `gh` CLI to create this issue for a given version?

klihub:
> Thanks [@bart0sh](https://github.com/bart0sh).
> 
> How do we feel about a script (e.g. `/hack/prepare-release.sh`) that uses the `gh` CLI to create this issue for a given version?

@elezar We could also set up an issue template for new releases, like this: https://github.com/cncf-tags/container-device-interface/pull/261

PR #260, opened 2025-03-07
Bump version to v1.0.0

bart0sh:
Ref: https://github.com/cncf-tags/container-device-interface/issues/259

### DIFF
--- a/cmd/cdi/go.mod
+++ b/cmd/cdi/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/spf13/cobra v1.6.0
 	gopkg.in/yaml.v2 v2.4.0
 	sigs.k8s.io/yaml v1.3.0
-	tags.cncf.io/container-device-interface v0.0.0
+	tags.cncf.io/container-device-interface v1.0.0
 	tags.cncf.io/container-device-interface/schema v0.0.0
 )
 
--- a/cmd/validate/go.mod
+++ b/cmd/validate/go.mod
@@ -11,7 +11,7 @@ require (
 	golang.org/x/mod v0.19.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
-	tags.cncf.io/container-device-interface v0.0.0 // indirect
+	tags.cncf.io/container-device-interface v1.0.0 // indirect
 	tags.cncf.io/container-device-interface/specs-go v1.0.0 // indirect
 )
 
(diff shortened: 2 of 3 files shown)


cmd/cdi/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/spf13/cobra v1.6.0
 	gopkg.in/yaml.v2 v2.4.0
 	sigs.k8s.io/yaml v1.3.0
-	tags.cncf.io/container-device-interface v0.0.0
+	tags.cncf.io/container-device-interface v1.0.0
```
elezar, line 12: More a comment on the release process: These are probably not strictly necessary since these CLIs are not generally consumed externally, but it's probably more consistent to have them.

cmd/cdi/go.mod
```diff
@@ -9,7 +9,7 @@ require (
...
 	tags.cncf.io/container-device-interface/schema v0.0.0
```
elezar, line 13: Question: Do we want to also specify a version for this?

elezar (approved):
Minor comments -- more related to the release process and not blockers.

cmd/cdi/go.mod
klihub: I was wondering the same. We anyway have the replace rule, so the version here has no effect, but if we bump it now, we should always go through the motions for consistency. @bart0sh So wouldn't leaving this at `0.0.0` be easier and of a burden for maintenance ?

klihub (approved):
I'm also fine with this, although I think it would be less of a maintenance burden to keep the in repo replaced version references always at `v0.0.0` and never touch them. Otherwise, we always need to remember to bump them even if this has absolutely no effect.

cmd/cdi/go.mod
bart0sh: done.

cmd/cdi/go.mod
bart0sh: reverted

bart0sh:
I've just realised that those changes were made by `make mod-tidy && make mod-verify`. Should I commit them again?

bart0sh:
Without them CI jobs failed.

cmd/cdi/go.mod
bart0sh: brougth both changes back as reverting them resulted in CI failures: https://github.com/cncf-tags/container-device-interface/actions/runs/13724737021

klihub:
> Without them CI jobs failed.

Ah, ok. Fair enough.

elezar:
> I've just realised that those changes were made by `make mod-tidy && make mod-verify`. Should I commit them again?

I think if the changes are made by tooling then that's fine. I don't want us to have to manually update things in those modules.

PR #275, opened 2025-06-19
Ensure that filemode is set for device nodes

elezar:
This change ensures that the filemode for a CDI device is properly set from the host device node when adding devices to the CDI spec.

Without this the device node permissions are 000 in the container under certain conditions (such as Kata).

See also https://github.com/NVIDIA/nvidia-container-toolkit/pull/960


elezar:
@zvonkok for the Kata use case the rust implementation must also be updated.

klihub (approved):

pkg/cdi/container-edits_unix.go
```diff
@@ -65,22 +86,31 @@ func (d *DeviceNode) fillMissingInfo() error {
...
 	}
-	if d.Major == 0 && d.Type != "p" {
-		d.Major = major
-		d.Minor = minor
+
+	if d.FileMode == nil {
+		d.FileMode = &di.fileMode
+	}
+
+	if d.Type == "p" {
```
bart0sh, line 107:
```suggestion
	if d.Type == fifoDevice {
```

pkg/cdi/container-edits_unix.go
```diff
@@ -65,22 +86,31 @@ func (d *DeviceNode) fillMissingInfo() error {
 		return nil
```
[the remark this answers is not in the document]
elezar: What would you set it to in this case?

pkg/cdi/container-edits_unix.go
bart0sh: Yes, please do.

klihub (approved):
@elezar @bart0sh This looks good to me. I only have one question related to a comment in the code about rootless containers.

PR #277, opened 2025-06-26
README: Docker 28.2 enables CDI by default

vvoland:
Docker 28.2 no longer requires an explicit opt-in to use CDI: https://github.com/moby/moby/pull/49963 Reflect that information in the README.

Also, since the `--device` behavior is the same for both Docker and Podman, combine them into one section.


elezar (approved):
Thanks for this. 

@klihub do you have any comments?

PR #278, opened 2025-08-06
Remove deprecated RDT fields

askervin:
Drop IntelRdt.EnableCMT and EnableMBM, and switch to EnableMonitoring to match upstream runtime-spec.


klihub:
> Ach, CDI had those fields defined 🤦
> 
> Yes, those were removed from the runtime spec as they were really problematic and had no known runtime implementation. See [opencontainers/runtime-spec#1287](https://github.com/opencontainers/runtime-spec/pull/1287)
> 
> Not sure how CDI should handle these. Just drop or filter out (or smth else) 🤔

I guess ideally/in principle we should deprecate those fields. So basically, remove the them from IntelRDT.toOCI() output, log a warning or an error if we load a CDI Spec file with either of those fields set to true, then in the next phase also refuse to load the whole CDI Spec if either is set. If we chose to do this, then additionally we should set the newly introduced `EnableMonitoring` field if either of the deprecated fields is set to true.

Then again, according to a comment by @marquiz in the above linked PR no known runtime implementations impemented either of these fields so if any CDI Spec currently in use tries to set those it anyway does not work correctly. So maybe we could just drop them altogether, like it is done in this PR. @elezar @bart0sh WDYT would be the proper course of actions here ?

klihub:
> @marquiz since you were the one to add the fields in the first place could you comment on whether these fields are used "in the wild"? If we think the risk of this is low we could drop the fields -- although this would required a spec version bump.
> 
> Also, how would we ensure that a modified spec is compatible with the _incoming_ OCI runtime spec version as we would not want to adjust this.

@elezar According to @marquiz' original comment https://github.com/opencontainers/runtime-spec/pull/1287#issue-3113305518, no known runtime implementation has ever implemented either of these fields. So it would seem to me that it should be safe to skip the deprecate-then-remove transitions and drop these in a single go.

marquiz:
> @marquiz since you were the one to add the fields in the first place could you comment on whether these fields are used "in the wild"? If we think the risk of this is low we could drop the fields -- although this would required a spec version bump.

I'd say the risk of somebody using those fields is very very low. There are no implementations for those and their specification was very vague/contradicting.

> I guess ideally/in principle we should deprecate those fields. So basically, remove the them from IntelRDT.toOCI() output, log a warning or an error if we load a CDI Spec file with either of those fields set to true, then in the next phase also refuse to load the whole CDI Spec if either is set.

This might be the most proper way.

> If we chose to do this, then additionally we should set the newly introduced `EnableMonitoring` field if either of the deprecated fields is set to true.

I would suggest not to do this to avoid further confusion and possible breakage in the future. People start to use and depend on these (for accident or whatever) and then if they're removed we brake them.

cmd/cdi/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/fsnotify/fsnotify v1.5.1
-	github.com/opencontainers/runtime-spec v1.1.0
+	github.com/opencontainers/runtime-spec v1.2.2-0.20250804081626-bfdffd548aa6
```
elezar, line 7: If the new field is not yet in a new released spec version, then we could first remove the deprecated fields and then add the new field once a tag has been released.

askervin:
@elezar, indeed, the new field is not yet released in runtime-spec. Removing old fields is not released either. So we are pretty well in sync.

Anyway, if handling this change in two PRs makes still sense to you, I'll keep this as "drop deprecated fields" PR, and create new PR that adds the new field. How that sounds?

elezar:
> > @elezar, indeed, the new field is not yet released in runtime-spec. Removing old fields is not released either. So we are pretty well in sync.
> > Anyway, if handling this change in two PRs makes still sense to you, I'll keep this as "drop deprecated fields" PR, and create new PR that adds the new field. How that sounds?
> 
> @askervin @elezar Then IMO at least the PR for the new field should be kept as a draft and updated once a new runtime spec is tagged which has the new field. Removing the deprecated field could be merged right away IMO, though (as long as it does not change our runtime-spec dependency).

Yes, let's drop the deprecated fields in a PR now (this should NOT have any dependency updates) and create a draft PR to be updated once a tagged spec is available for the new field.

klihub (approved):

elezar (approved):
lgtm

PR #288, opened 2025-10-18
cdi: inject mount UID/GID mappings if user NS is in use.

### LINKED ISSUE #287: [https://github.com/kubernetes/kubernetes/issues/134604] CDI-injected bind-mount inaccessible for pods with `hostUsers: false`
https://github.com/kubernetes/kubernetes/issues/134604 reported against kubernetes proper is actually a problem with CDI. See the original issue for details and discussion.

As a short summary, when a pod is running with `hostUsers: false`, then all mounts should (and kubelet-requested runtime-injected ones will) have UID- and GID-mappings. Since CDI inject mounts without any UID- and GID-mappings containers trying to access CDI-injected bind-mounted directories will fail with `EPERM` unless the mounted directory has at least `0666` permissions set.

klihub:
When injecting mounts to a container with user namespaces in use, inject the mounts with UID and GID mappings taken from the OCI Spec.

Fixes #287.

@elezar @haircommander Although this simplistic fix works for the reported case, it might be a bit too simplistic (in the long run).

In particular, I wonder if we should eventually provide a way for the one requesting CDI injection (typically a DRA driver) to also provide UID-/GID-mappings for injected mounts, and only let CDI itself pick and inject UID-/GID-mappings on its own, if the necessary preconditions apply but mappings are not provided.

 IOW, should we provide a way for a DRA driver itself to pick and inject the mappings ? If we are of the opinion that we should, then I think we have a bit of a problem with APIs and how things are currently wired up with CDI/CRI. The straightforward way to allow a driver to pick mappings would be to simply extend the CDI notion of a mount with UID- and GID-mappings. However, that would imply that only dynamically generated CDI Specs would allow injecting non-static mappings, since there is no way to pass on any additionald/dynamic parameters along a CDI device injection request over CRI (or in the CDI/runtime API).

### DIFF
--- a/pkg/cdi/oci.go
+++ b/pkg/cdi/oci.go
@@ -30,14 +30,46 @@ func (h *Hook) toOCI() spec.Hook {
 	}
 }
 
+// Additional OCI mount option to apply to injected mounts.
+type ociMountOption func(*spec.Mount)
+
+// withIDMapForBindMount adds any necessary ID mapping options for a bind mount.
+func withIDMapForBindMount() ociMountOption {
+	return func(m *spec.Mount) {
+		option := ""
+		if m.Type == "bind" {
+			option = "idmap"
+		}
+
+		for _, o := range m.Options {
+			switch o {
+			case "idmap", "ridmap":
+				return
+			case "bind":
+				option = "idmap"
+			case "rbind":
+				option = "ridmap"
+			}
+		}
+
+		if option != "" {
+			m.Options = append(m.Options, option)
+		}
+	}
+}
+
 // toOCI returns the opencontainers runtime Spec Mount for this Mount.
-func (m *Mount) toOCI() spec.Mount {
-	return spec.Mount{
+func (m *Mount) toOCI(options ...ociMountOption) spec.Mount {
+	om := spec.Mount{
 		Source:      m.HostPath,
 		Destination: m.ContainerPath,
 		Options:     m.Options,
 		Type:        m.Type,
 	}
+	for _, o := range options {
+		o(&om)
+	}
+	return om
 }
 
 // toOCI returns the opencontainers runtime Spec LinuxDevice for this DeviceNode.
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -123,8 +123,15 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 
 	if len(e.Mounts) > 0 {
 		for _, m := range e.Mounts {
+			mnt := &Mount{m}
+
 			specgen.RemoveMount(m.ContainerPath)
-			specgen.AddMount((&Mount{m}).toOCI())
+
+			if !specHasUserNamespace(spec) {
+				specgen.AddMount(mnt.toOCI())
+			} else {
+				specgen.AddMount(mnt.toOCI(withIDMapForBindMount()))
+			}
 		}
 		sortMounts(&specgen)
 	}
@@ -464,4 +471,17 @@ func (m orderedMounts) Swap(i, j int) {
 // parts returns the number of parts in the destination of a mount. Used in sorting.
 func (m orderedMounts) parts(i int) int {
 	return strings.Count(filepath.Clean(m[i].Destination), string(os.PathSeparator))
+}
+
+// specHasUserNamespace returns true if the OCI Spec has a Linux UserNamespace.
+func specHasUserNamespace(spec *oci.Spec) bool {
+	if spec == nil || spec.Linux == nil {
+		return false
+	}
+	for _, ns := range spec.Linux.Namespaces {
+		if ns.Type == oci.UserNamespace {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/cdi/container-edits_test.go
+++ b/pkg/cdi/container-edits_test.go
@@ -800,6 +800,104 @@ func TestApplyContainerEdits(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "mount added to container with Linux user namespace and uid/gid mappings",
+			spec: &oci.Spec{
+				Linux: &oci.Linux{
+					UIDMappings: []oci.LinuxIDMapping{
+						{
+							ContainerID: 0,
+							HostID:      1000,
+							Size:        999,
+						},
+					},
+					GIDMappings: []oci.LinuxIDMapping{
+						{
+							ContainerID: 0,
+							HostID:      2000,
+							Size:        777,
+						},
+					},
+					Namespaces: []oci.LinuxNamespace{
+						{
+							Type: oci.UserNamespace,
+							Path: "/foo/bar",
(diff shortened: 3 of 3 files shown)


pkg/cdi/oci.go
```diff
@@ -30,14 +30,33 @@ func (h *Hook) toOCI() spec.Hook {
 	}
 }
 
+// Extra OCI mount option to apply to injected mounts.
+type extraOCIMountOption func(*spec.Mount)
+
+// withMountIDMappings adds UID and GID mappings for the given mount.
+func withMountIDMappings(uid, gid []spec.LinuxIDMapping) extraOCIMountOption {
```
elezar: Does it make sense to have `uid` and `gid` settable by separate options.

klihub: I am fine with anything. Separate options, combined options, even with simply doing it like this

```golang
func (m *Mount) toOCI(uids, gids []spec.LinuxIDMapping) spec.Mount {
	om := spec.Mount{
		Source:      m.HostPath,
		Destination: m.ContainerPath,
		Options:     m.Options,
		Type:        m.Type,
		UIDMappings: uids,
		GIDMappings: gids,
	}
	return om
```

klihub:
> Is something similar required for devices since the uid and gid in the spec are the IDs in the container (https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices)?

At least nothing identical we can do for devices, since they don't have UID/GID mappings, only mounts have them. Also, device access control is via permissions and cgroup access rules, AFAIK.

rata:
@elezar sure.

The devices object in the runtime-spec has a uid/gid: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices. I'm proposing to set those when devices are injected.

I'm proposing to set it to the user mapped inside the userns. The mapping in the OCI runtime spec is here: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#user-namespace-mappings. Using the example from the spec:
```
"uidMappings": [
    {
        "containerID": 0,
        "hostID": 1000,
        "size": 32000
    }
],
"gidMappings": [
    {
        "containerID": 0,
        "hostID": 1000,
        "size": 32000
    }
]
```

If the device we want (https://github.com/cncf-tags/container-device-interface/blob/ffd52dc97d2d71c1b50ed966aba1eaa016c70412/specs-go/config.go#L49-L54) the hostPath is owned by UID 0 and GID 0, then using this examples we want to assign it to "0 inside the container". Then, we check what is the mapped ID for 0 inside the container.

The UID and GID 0 inside the container are mapped to the ID 1000 outside the container. Then, the device will have the uid and gid set to 1000 in this example.

Is this example clear? Don't hesitate to ask otherwise :)

rata:
@everzakov ohh, thanks for catching this! Okay, so runc is already setting the uid/gid to root inside the container and that doesn't work :(

So, for bind-mounts of filesystems (a file or a directory), what this PR is doing is fine.

For other devices, we need to see what we can do. Because the option that should work, isn't working and as @everzakov was saying maybe it can't work without changing the uid/gid of the device on the host, that might be problematic.

everzakov:
> Ping @everzakov @rata. What do you guys think ? Should we take this in ?

Sorry for a late answer. Can i try this change with mount which does not support mount_setattr idmap? This feature is [required](https://github.com/opencontainers/runc/blob/main/libcontainer/mount_linux.go#L316) by runc. I want to check if the error will be transparent.

rata:
@klihub sorry, I was away. Yes, sorry if I wasn't clear [here][link]. What I wanted to say is: doing what this PR does for bind-mounts is what it should be done.

If the container runtime doesn't support user-namespaces, then the kubelet will fail the pod before this. If it does, then we can just add these fields safely. So this is sound, as long as it is done only for bind-mounts.


[link]: https://github.com/cncf-tags/container-device-interface/pull/288#issuecomment-3502468944

rata (commented):
This mostly LGTM. I think this is probably fine if it works, but left some comments to make it less risky2

elezar (commented):
This looks quite a bit cleaner. My one question is whether we really need both `hasUserNamespace` and `isBind` to control this behaviour.

Would applying the following diff make things simpler / more self-contained:
```
diff --git a/pkg/cdi/container-edits.go b/pkg/cdi/container-edits.go
index 08be630..848602c 100644
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -124,21 +124,16 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 	if len(e.Mounts) > 0 {
 		var (
 			hasUserNamespace = specHasUserNamespace(spec)
-			isBind           bool
-			bindKind         string
 		)
 		for _, m := range e.Mounts {
 			mnt := &Mount{m}
-			if hasUserNamespace {
-				isBind, bindKind = mnt.isBindMount()
-			}
 
 			specgen.RemoveMount(m.ContainerPath)
 
-			if !hasUserNamespace || !isBind {
-				specgen.AddMount(mnt.toOCI())
+			if hasUserNamespace {
+				specgen.AddMount(mnt.toOCI(withMountIDMappings()))
 			} else {
-				specgen.AddMount(mnt.toOCI(withMountIDMappingsFor(bindKind)))
+				specgen.AddMount(mnt.toOCI())
 			}
 		}
 		sortMounts(&specgen)
diff --git a/pkg/cdi/oci.go b/pkg/cdi/oci.go
index 1071e12..5ce35cb 100644
--- a/pkg/cdi/oci.go
+++ b/pkg/cdi/oci.go
@@ -33,9 +33,22 @@ func (h *Hook) toOCI() spec.Hook {
 // Additional OCI mount option to apply to injected mounts.
 type ociMountOption func(*spec.Mount)
 
-// withMountIDMappingsFor adds the necessary ID mapping options for a bind mount.
-func withMountIDMappingsFor(kind string) ociMountOption {
+// withMountIDMappings adds the necessary ID mapping options for a bind mount.
+func withMountIDMappings() ociMountOption {
 	return func(m *spec.Mount) {
+		var kind string
+		for _, o := range m.Options {
+			switch o {
+			case "rbind":
+				kind = "rbind"
+				break
+			case "bind":
+				kind = "bind"
+			}
+		}
+		if kind == "" {
+			kind = m.Type
+		}
 		switch kind {
 		case "rbind":
 			m.Options = append(m.Options, "ridmap")

```

rata (approved):
LGTM modulo the lack of specifying mappings that I mentioned before.

rata:
I mean this: https://github.com/cncf-tags/container-device-interface/pull/288#discussion_r2578477281

You asked me for review, that is my comment. Other than that, LGTM :)

rata:
Friendly ping? IMHO this is ready to merge. We are aiming for userns GA in the next kube release, it will be great if this fix is in :)

elezar (approved):
THanks @klihub.

I think this looks much simpler now.

kad (approved):
LGTM

PR #290, opened 2025-11-14
go.{mod,sum}: bump runtime tools, spec deps to v1.3.0.

klihub:
Bump OCI runtime tools and spec to v1.3.0.


elezar (approved):
Thanks @klihub.

Is this something that we need to "backport" to the v1.0.0 spec, or are the changes on `main` such that we can go ahead and just release `v1.0.1`?

PR #294, opened 2025-11-20
specs-go: fix minumum version check for IntelRdt.

klihub:
This slipped through our review...

Add minimum version check unit test for IntelRdt. Fix check to require `1.1.0` if at least one of the global or device specific container edits use IntelRdt.

### DIFF
--- a/specs-go/version.go
+++ b/specs-go/version.go
@@ -149,6 +149,15 @@ func requiresV110(spec *Spec) bool {
 			return true
 		}
 	}
+
+	for _, dev := range spec.Devices {
+		if i := dev.ContainerEdits.IntelRdt; i != nil {
+			if i.Schemata != nil || i.EnableMonitoring {
+				return true
+			}
+		}
+	}
+
 	return false
 }
 


cmd/cdi/cmd/validate.go
```diff
@@ -49,6 +50,15 @@ were reported by the cache.`,
 				fmt.Printf("  %2d: %v\n", idx, strings.TrimSpace(err.Error()))
 			}
 		}
+
+		for _, v := range cache.ListVendors() {
+			for _, s := range cache.GetVendorSpecs(v) {
```
bart0sh: I can see this nested loop in other places as well. Would it make sense to wrap it in function?

bart0sh (approved):
/lgtm

specs-go/version.go
```diff
@@ -149,6 +149,15 @@ func requiresV110(spec *Spec) bool {
 			return true
 		}
 	}
+
+	for _, dev := range spec.Devices {
```
elezar: Good catch @klihub. Thanks.

PR #301, opened 2025-12-16
cdi,SPECS.md: allow empty cgroup permissions.

### LINKED ISSUE #300: CDI spec does not allow empty permissions for device nodes
When creating a container it can make sense to not allow any operation at first on specified devices: the device node will still be created and the associated shared libraries and tools can still be mounted, eg everything will be prepared for a later usage but the device won't be 'open' yet

This use case is close to the one nvidia-container-toolkit supported in legacy mode with the no-cgroups = true flag (at least this was the way I was using it, I may have misunderstood its purpose, cc @elezar here, will probably be able to clarify :) )

But in the spec, it's explicitely specified 

```
// Cgroups permissions of the device, candidates are one or more of [... r, w, m]
```

And we can indeed see in the implem here that an empty string is translated to "rwm"
https://github.com/cncf-tags/container-device-interface/blob/98a7d735d9592c1ccca0130e06b0709f4b65ecbf/pkg/cdi/container-edits.go#L109

Proposal: change this spec to  
```
// Cgroups permissions of the device, candidates are an empty string or any combination of one or more of [... r, w, m] 
```

What do you think ?

side note: so far, containerd users are lucky because containerd seems quite flexible and discards unknown chars in the allow string of the oci spec. This statement, coupled with the fact that the cdi spec implem above is flexible enough to pass through any invalid permission string to the oci spec without raising any error, setting the allow string to any unknown char of the cdi spec luckily creates the intented effect, but this is more a flaw than a spec respect, if I understand correctly

klihub:
Allow injecting devices with empty cgroup permissions, requested by the `"none"` permission string or the new `pkg/cdi.NoPermissions` constant.

Fixes #300

### DIFF
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -42,6 +42,9 @@ const (
 	PoststartHook = "poststart"
 	// PoststopHook is the name of the OCI "poststop" hook.
 	PoststopHook = "poststop"
+
+	// NoPermissions requests empty cgroup permissions for a device.
+	NoPermissions = "none"
 )
 
 var (
@@ -106,8 +109,11 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 
 		if dev.Type == "b" || dev.Type == "c" {
 			access := d.Permissions
-			if access == "" {
+			switch access {
+			case "":
 				access = "rwm"
+			case NoPermissions:
+				access = ""
 			}
 			specgen.AddLinuxResourcesDevice(true, dev.Type, &dev.Major, &dev.Minor, access)
 		}
@@ -361,12 +367,14 @@ func (d *DeviceNode) Validate() error {
 	if _, ok := validTypes[d.Type]; !ok {
 		return fmt.Errorf("device %q: invalid type %q", d.Path, d.Type)
 	}
-	for _, bit := range d.Permissions {
-		if bit != 'r' && bit != 'w' && bit != 'm' {
-			return fmt.Errorf("device %q: invalid permissions %q",
-				d.Path, d.Permissions)
-		}
+	switch {
+	case d.Permissions == "":
+	case d.Permissions == NoPermissions:
+	case strings.Trim(d.Permissions, "rwm") != "":
+		return fmt.Errorf("device %q: invalid permissions %q",
+			d.Path, d.Permissions)
 	}
+
 	return nil
 }
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -139,6 +139,7 @@ The keywords "must", "must not", "required", "shall", "shall not", "should", "sh
                     // * r - allows container to read from the specified device.
                     // * w - allows container to write to the specified device.
                     // * m - allows container to create device files that do not yet exist.
+                    // Omitted or empty permissions default to 'rwm'. 'none' requests empty permissions.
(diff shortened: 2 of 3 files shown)


kad (approved):
looks ok to me.

pkg/cdi/container-edits.go
```diff
@@ -354,10 +360,12 @@ func (d *DeviceNode) Validate() error {
...
-	for _, bit := range d.Permissions {
-		if bit != 'r' && bit != 'w' && bit != 'm' {
-			return fmt.Errorf("device %q: invalid permissions %q",
-				d.Path, d.Permissions)
+	if d.Permissions != NoPermissions {
+		for _, bit := range d.Permissions {
+			if bit != 'r' && bit != 'w' && bit != 'm' {
+				return fmt.Errorf("device %q: invalid permissions %q",
+					d.Path, d.Permissions)
+			}
```
elezar: Does this check cover the cases that we're interested in? The way I understand it, valid values for `DeviceNode.Permissions` are now `"none"`, `""`, and combinations of `"rwm"`.  Does that mean that we need to implement a function like:
```
func AssertValidDevicePermissions(p string) error {
    switch {
    case p == "":
        return nil
    case p == NoPermissions: {
        return nil
    case strings.Trim(p, "rwm") == "":
       return nil
    default:
        return fmt.Errorf("invalid permission %v", p)
    }
}
```

elezar (dismissed):
The addition of `none` as a special case looks good. I do think that the device validation needs to be properly updated for this though.

pkg/cdi/container-edits.go
klihub: Sure, updated to use a switch for permission validation and added the 'notnone' permission test case as suggested.

pkg/cdi/container-edits.go
```diff
@@ -106,8 +109,11 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 
 		if dev.Type == "b" || dev.Type == "c" {
 			access := d.Permissions
-			if access == "" {
+			switch access {
+			case "":
 				access = "rwm"
```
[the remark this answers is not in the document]
elezar: Yes, this is covered in one of the `TestApplyContainerEdits` tests. For example:
```
		{
			name: "empty spec, device",
			spec: &oci.Spec{},
			edits: &cdi.ContainerEdits{
				DeviceNodes: []*cdi.DeviceNode{
					{
						Path: "/dev/null",
					},
				},
			},
			result: &oci.Spec{
				Linux: &oci.Linux{
					Devices: []oci.LinuxDevice{
						{
							Path:     "/dev/null",
							Type:     "c",
							Major:    nullDeviceMajor,
							Minor:    nullDeviceMinor,
							FileMode: nullDeviceFileMode,
						},
					},
					Resources: &oci.LinuxResources{
						Devices: []oci.LinuxDeviceCgroup{
							{
								Allow:  true,
								Type:   "c",
								Major:  &nullDeviceMajor,
								Minor:  &nullDeviceMinor,
								Access: "rwm",
							},
						},
					},
				},
			},
		},

bart0sh (approved):
/lgtm

bart0sh:
/assign @elezar 
for approval & merge

elezar (approved):

PR #313, opened 2026-04-16
Makefile: add GO_EXTRAFLAGS to go build call

koenkooi:
We may need to specify things like -trimpath, so add $(GO_EXTRAFLAGS) to allow them to be passed from the buildsystem to the Makefile.

This is especially true for reproducible builds in both Debian and OpenEmbedded, golang is very fond of embedding buildpaths into the binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ golangci-lint ci-lint:
 
 $(BINARIES): bin/%:
 	$(Q)echo "Building $@..."
-	$(Q)(cd cmd/$(*) && $(GO_BUILD) -o $(abspath $@) .)
+	$(Q)(cd cmd/$(*) && $(GO_BUILD) $(GO_EXTRAFLAGS) -o $(abspath $@) .)
 
 #
 # go module tidy and verify targets


elezar (commented):
I don't really have an objection to adding these flags, but just wanted to note that the cmds included in this repo are for testing and experimentation purposes only. Are you using them in an environment where reproducibility is a requirement?

klihub (approved):
LGTM, but with the fair warning @elezar has already stated: these binaries are considered test tools. This is mostly reflected in them not being regularly tested extensively. It's good to be aware of that.

koenkooi:
> I don't really have an objection to adding these flags, but just wanted to note that the cmds included in this repo are for testing and experimentation purposes only.

I fully realize this, but I do want to create packages for this to integrate this nicely in our testing framework. The fewer patches I need locally, the better :)

> Are you using them in an environment where reproducibility is a requirement?

The OpenEmbedded build system errors out when it find references to the buildpath in the binary. Adding `-trimpath` fixes that. There are other flags that I want passed, but trimpath was the catalyst for this patch.

The runtime environment doesn't care about reproducibility, this is for debugging and validating the output from a CDI generator: https://github.com/qualcomm-linux/qualcomm-CDI-generator/

elezar (approved):
Lgtm

Issue #302, opened 2025-12-23
[Extension] Add way to configure the allowed devices list


kc9jud:
Currently there is no way to add entries to the [allowed device list](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#allowed-device-list). This means that there is no way to specify a wildcard entry, for devices that may be created during the lifetime of a container. Adding this would allow functionality equivalent to Docker's [`--device-cgroup-rule`](https://docs.docker.com/reference/cli/docker/container/run/#device-cgroup-rule).

elezar:
Let's keep this open while we review 314.

PR #314, opened 2026-04-17
Add support for cgroup-only device nodes

elezar:
This change proposes a mechanism to allow for the specification of cgroup-only device nodes (See #302). This is represented by a device node with a `path` equal to `*`. For such a device node, the device node information is NOT injected into the OCI runtime spec, but a cgroup rule is added for the major and minor number of the device. This is similar to the Docker [`--device-cgroup-rule`](https://docs.docker.com/reference/cli/docker/container/run/#device-cgroup-rule) argument and means that the device node NEED not exist on the host.

In addition, a mojor or minor number of `-1` as specified in `opencontainers/cgroup` https://github.com/opencontainers/cgroups/blob/d492e44fbd24c97075c46c33e67ae43b5c7fa0ac/devices/config/device.go#L10 and used in `runc` is treated as a "wildcard" device matching all major (or minor) numbers.

Note that `runc` skips device nodes with an empty path: https://github.com/opencontainers/runc/blob/edbed618bff99edaaa9d622f9f8687004fe5e50a/libcontainer/rootfs_linux.go#L984-L988 assuming they are only provided for cgroup access.

### DIFF
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -374,19 +374,21 @@ func (d *DeviceNode) addToGenerator(specgen *ocigen.Generator, spec *oci.Spec) e
 	specgen.AddDevice(dev)
 
 	if dev.Type == "b" || dev.Type == "c" {
-		access := d.Permissions
-		switch access {
-		case "":
-			access = "rwm"
-		case NoPermissions:
-			access = ""
-		}
-		specgen.AddLinuxResourcesDevice(true, dev.Type, &dev.Major, &dev.Minor, access)
+		specgen.AddLinuxResourcesDevice(true, dev.Type, &dev.Major, &dev.Minor, d.getAccessString())
 	}
 	return nil
 }
 
+func (d *DeviceNode) getAccessString() string {
+	switch d.Permissions {
+	case "":
+		return "rwm"
+	case NoPermissions:
+		return ""
+	default:
+		return d.Permissions
 	}
+}
 	return nil
 }
 // Hook is a CDI Spec Hook wrapper, used for validating hooks.
--- a/pkg/cdi/container-edits.go
+++ b/pkg/cdi/container-edits.go
@@ -328,19 +328,11 @@ type DeviceNode struct {
 
 // Validate a CDI Spec DeviceNode.
 func (d *DeviceNode) Validate() error {
-	validTypes := map[string]struct{}{
-		"":  {},
-		"b": {},
-		"c": {},
-		"u": {},
-		"p": {},
-	}
-
 	if d.Path == "" {
 		return errors.New("invalid (empty) device path")
 	}
-	if _, ok := validTypes[d.Type]; !ok {
-		return fmt.Errorf("device %q: invalid type %q", d.Path, d.Type)
+	if strings.Contains(d.Path, "*") && d.Path != "*" {
+		return errors.New("invalid device device path; wildcard device path")
 	}
 	switch {
 	case d.Permissions == "":
@@ -350,10 +342,41 @@ func (d *DeviceNode) Validate() error {
 			d.Path, d.Permissions)
 	}
 
+	switch {
+	case d.isWildcardDevice():
+		validTypes := map[string]struct{}{"b": {}, "c": {}}
+		if _, ok := validTypes[d.Type]; !ok {
+			return fmt.Errorf("device %q: invalid type %q", d.Path, d.Type)
+		}
+	default:
+		validTypes := map[string]struct{}{
+			"":  {},
+			"b": {},
+			"c": {},
+			"u": {},
+			"p": {},
+		}
+		if _, ok := validTypes[d.Type]; !ok {
+			return fmt.Errorf("device %q: invalid type %q", d.Path, d.Type)
+		}
+		// For a regular device, we also check that we have valid major and
+		// minor numbers.
+		if d.Major == -1 {
+			return fmt.Errorf("wildcard device major requires a wildcard device")
+		}
+		if d.Minor == -1 {
+			return fmt.Errorf("wildcard device minor requires a wildcard device")
+		}
+	}
+
 	return nil
 }
 
 func (d *DeviceNode) addToGenerator(specgen *ocigen.Generator, spec *oci.Spec) error {
+	if d.isWildcardDevice() {
+		return d.addAsWildcardDevice(specgen)
+	}
+
 	err := d.fillMissingInfo()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


pkg/cdi/container-edits.go
```diff
@@ -378,6 +353,77 @@ func (d *DeviceNode) Validate() error {
...
+		return fmt.Errorf("wildcard device node supported for device type %v", d.Type)
+	}
+	var major *int64
+	if d.Major != -1 {
+		major = &d.Major
+	}
+	var minor *int64
+	if d.Minor != -1 {
+		minor = &d.Minor
+	}
```
elezar, line 421: It feels as if this is the CORE change here since this SHOULD allow us to inject the cgroup rule even without performing additional checks as long as SOME non-zero device major is specified. This works because we explicitly ignore errors getting device node information if at least the Major and Type is is specified. What we need is the `-1` to `nil` mapping since `nil` is interpreted as `*` by the OCI runtime spec.

pkg/cdi/container-edits.go
```diff
@@ -378,6 +353,77 @@ func (d *DeviceNode) Validate() error {
...
+// Such devices do not cause device nodes to be created in the container, but
+// do update the cgroups to allow device access.
+// A wildcard device in the CDI spec has a major and/or minor number set to -1
+// (as is also the convention in runc) OR includes a "*" in the path.
+func (d *DeviceNode) isWildcardDevice() bool {
+	if d.Minor == -1 || d.Major == -1 {
+		return true
+	}
+	return strings.Contains(d.Path, "*")
+}
```
elezar, line 408: An alternative to reusing the `DeviceNode` struct is to add an explicity spec type for cgroups. That does complicate the spec a bit since there is an implied connection between the `DeviceNodes` and the cgroups added to the container.

pkg/cdi/container-edits.go
```diff
@@ -378,6 +353,77 @@ func (d *DeviceNode) Validate() error {
...
+
+func (d *DeviceNode) addAsWildcardDevice(specgen *ocigen.Generator) error {
+	if d.Type != "b" && d.Type != "d" {
+		return fmt.Errorf("wildcard device node supported for device type %v", d.Type)
```
klihub, line 412: Should this error be "... not supported for..." ?

pkg/cdi/container-edits.go
```diff
@@ -378,6 +353,77 @@ func (d *DeviceNode) Validate() error {
...
+// (as is also the convention in runc) OR includes a "*" in the path.
+func (d *DeviceNode) isWildcardDevice() bool {
+	if d.Minor == -1 || d.Major == -1 {
+		return true
+	}
+	return strings.Contains(d.Path, "*")
+}
+
+func (d *DeviceNode) addAsWildcardDevice(specgen *ocigen.Generator) error {
+	if d.Type != "b" && d.Type != "d" {
```
klihub, line 411: Can the type really be "d" ? I think that would fail `DeviceNode.Validate()`.

pkg/cdi/container-edits.go
```diff
@@ -378,6 +353,77 @@ func (d *DeviceNode) Validate() error {
...
+// isWildcardDevice returns whether the device node represents a "wildcard"
+// device.
+// Such devices do not cause device nodes to be created in the container, but
+// do update the cgroups to allow device access.
+// A wildcard device in the CDI spec has a major and/or minor number set to -1
+// (as is also the convention in runc) OR includes a "*" in the path.
+func (d *DeviceNode) isWildcardDevice() bool {
+	if d.Minor == -1 || d.Major == -1 {
+		return true
+	}
```
klihub, line 406: Cannot we (and if we can, shouldn't we) always require that the device path is always set to "*" for such a device ? I think it would be less confusing if we didn't allow it, and doing it otherwise also might give the false impression that something might be filled in from the same device on the host...

klihub:
@elezar This looks fine to me as a feature to have, but I have a few questions about some implementation/design details.

pkg/cdi/container-edits.go
elezar: Yes ... it should be `c`. Let me update.

pkg/cdi/container-edits.go
elezar: My one thought here was whether we wanted to use the path for naming (from a user perspective), but maybe always using `*` is better. That is to say that a wildcard device ALWAYS has a path that is *exactly equal* to `"*"` and MAY include specific major and minor numbers or `-1` as a wildcard major or minor number.

pkg/cdi/container-edits.go
klihub: Just a mental note... It's a pity we don't have `Major` and `Minor` as `*int64` instead, so we could differentiate between zero- and omitted values.

klihub (approved):
This LGTM now.

klihub (commented):
This LGTM now. We'll need to update documentation and tests, though.

klihub (changes requested):

elezar:
> This LGTM now. We'll need to update documentation and tests, though.

Yes, I just wanted to get some early feedback before adding more of those.

Another question. I suppose that this would justify a spec bump, or would we consider back-porting this to older versions? If we do bump the spec, then we would have to update the minimum version checks.

pkg/cdi/container-edits.go
klihub: I think it would be better to avoid such a complication. I think your current implementation is a better option.

klihub:
> > This LGTM now. We'll need to update documentation and tests, though.
> 
> Yes, I just wanted to get some early feedback before adding more of those.
> 
> Another question. I suppose that this would justify a spec bump, or would we consider back-porting this to older versions? If we do bump the spec, then we would have to update the minimum version checks.

Yes, I think this probably warrants a spec version bump. In principle it changes how we semantically interpret the spec. Although now that we require the device name for a wildcard injection to be a literal "*", the practical difference is reduced to a "it would have failed earlier but now it won't and will have a special meaning", and it hard to imagine that someone could possibly rely on the old failing behavior. But better be safe.

About backporting (without bumping the Spec version), I think that would not help those who need the new behavior, if/when they ran with an already released runtime version. That would just unexpectedly fail.

pkg/cdi/container-edits.go
```diff
@@ -375,6 +342,101 @@ func (d *DeviceNode) Validate() error {
...
+			"":  {},
+			"b": {},
+			"c": {},
+			"u": {},
+			"p": {},
+		}
+		if _, ok := validTypes[d.Type]; !ok {
+			return fmt.Errorf("device %q: invalid type %q", d.Path, d.Type)
+		}
+		// For a regular device, we also check that
```
klihub: This first line looks like the beginning of an old comment.

elezar: Updated.

specs-go/version.go
```diff
@@ -60,6 +61,7 @@ var validSpecVersions = requiredVersionMap{
 	v080: requiresV080,
 	v100: requiresV100,
 	v110: requiresV110,
+	v120: requiresV120,
```
elezar, line 64: Note to self. I still need to add tests for this.

pkg/cdi/container-edits.go
kad: I actually wonder, do we need "safety net' to prevent accidental "allow all devices" setting?

pkg/cdi/container-edits.go
elezar: cc @marquiz here we proposed reusing the DeviceNode type. Does that mean we should iterate on this PR here instead of #363?

marquiz: IMHO the wildcard device nodes and cgroup rules are rather orthogonal features and would be cleaner to not be mixed.

How I read/think about it:

1. Wildcard devices allows one to create all devices that match a glob pattern on the host. This will not silently/implicitly give access to any other devices than those "mirrored" from the host. 
2. Extra cgroups gives the permission to access devices (without creating any device nodes). Cleaner, explicit API, now "shadow" fields like path, hostpath, uid/gid etc involved.

IOW, in my view the DeviceNode type should be used for wildcard device nodes. A separate field would be better for wildcard device cgroup rules.

Not sure how others see it.

elezar: @marquiz this PR is actually not about wildcard devices (like #362). The `*` is just a indcator that we're adding a cgroup rule instead. There is no globbing being done and the mechanics are the same as your proposal. If we feel that a separate containerEdit type for cgroups is a better match here, then that is fine too.

PR #316, opened 2026-06-02
Remove dependency on `github.com/opencontainers/runtime-tools`

tianon-sso:
The `generate` package was used in two places purely as a convenience wrapper over `*oci.Spec` field mutations -- none of its logic was load-bearing. Replace it with direct spec manipulation throughout.

In `pkg/cdi/container-edits.go`, drop the `ocigen.Generator` value and operate on `*oci.Spec` directly in `Apply`. Extract small private helpers for the operations that have non-trivial logic: `addMultipleProcessEnv` (deduplicates by key), `removeDevice` and `removeMount` (find-and-splice), and `addProcessAdditionalGid` (append-if-absent).

In `cmd/cdi/cmd/cdi-api.go`, replace the three-step `gen.NewFromSpec` / `ClearLinuxDevices` / `AddDevice` sequence in `collectCDIDevicesFromOCISpec` with a single partition loop over `spec.Linux.Devices`.

Assisted-By: "claude my eyes right out"


tianon-sso:
Friendly ping - `runtime-tools` has historically been a problematic dependency across the ecosystem, and the lift to remove it is pretty small here :bow:

elezar:
Sorry for the delay here. I'll have a look tomorrow.

elezar (commented):
A first pass of this looks good. One question I would have is whether we wanted to add a local type that wrapps `*oci.Spec` like `runtime-tools` does to reduce the diff further, but I assume that it probably isn't worth the effort for the reduced feature set we have here.

@klihub what do you think?

pkg/cdi/container-edits.go
```diff
@@ -115,80 +119,134 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
...
 	}
 
 	return nil
 }
 
-func specgenAddLinuxNetDevice(specgen *ocigen.Generator, hostIf string, netDev *oci.LinuxNetDevice) {
-	if specgen == nil || netDev == nil {
+// addMultipleProcessEnv adds or replaces environment variables on the process,
+// deduplicating by key.
+func addMultipleProcessEnv(process *oci.Process, envs []string) {
```
elezar, line 198: One thing that I want to call out (thanks Codex). There is a subtle difference between the new code and the original `runtime-tools` implementation. In the previous implemenation, the entries in [`envCache`](https://github.com/opencontainers/runtime-tools/blob/8a4db579f5c88af5a0d036fad34bddc9c1f703f3/generate/generate.go#L545) were constucted using `ENV=value` strings when loading an OCI spec. This means that when updating envs by name from the CDI spec, we would not overwrite the initial envvar, but always append. (Subsequent CDI updates would append). This was likely a bug in `runtime-tools`.

Note that there is also a slight mismatch to the wording of the CDI specification which states that envvars are appended to the environment.

I do, however, think that the new behaviour is correct and should also remove ambiguity around dealing with multiple envvars. We may want to update the wording around evvars for the spec though.

pkg/cdi/container-edits.go
```diff
@@ -115,80 +119,134 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
...
+	for _, env := range envs {
+		parts := strings.SplitN(env, "=", 2)
+		if len(parts) < 2 {
+			continue
+		}
+		prefix := parts[0] + "="
+		replaced := false
+		for i, e := range process.Env {
```
elezar, line 206: Implementing a simplified generator type locally will allow us to cache the envvars to improve performance for containers with many envvars. (Using envvars in CDI is not common, so it may not be worth it though).

pkg/cdi/container-edits.go
```diff
@@ -115,80 +119,134 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
...
+			if strings.HasPrefix(e, prefix) {
+				process.Env[i] = env
+				replaced = true
+				break
```
elezar: not critical: Using a labeled continue would remove the need for an auxiliarry variable. Alternatively, we could add an `addEnvToProcess(process *oci.Process, env string)` function:
```
func addEnvToProcess(process *oci.Process, env string) {
	parts := strings.SplitN(env, "=", 2)
	if len(parts) < 2 {
		return
	}
	prefix := parts[0] + "="
	for i, e := range process.Env {
		if strings.HasPrefix(e, prefix) {
			process.Env[i] = env
			return
		}
	}
	
	process.Env = append(process.Env, env)
}
```

pkg/cdi/container-edits.go
```diff
@@ -104,8 +105,11 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 			}
 		}
 
-		specgen.RemoveDevice(dev.Path)
-		specgen.AddDevice(dev)
+		if spec.Linux == nil {
+			spec.Linux = &oci.Linux{}
+		}
+		removeDevice(spec, dev.Path)
+		spec.Linux.Devices = append(spec.Linux.Devices, dev)
```
elezar, line 112: Another inconsistency is that `specgen.AddDevice(dev)` would skip the addition if the spec had multiple matching devices to start with.

pkg/cdi/container-edits.go
tianon-sso: I updated with a named loop for now in a66c1dd but I'm happy to adjust to the singular form with a loop in the caller (or keeping both functions) instead if you prefer 👍

pkg/cdi/container-edits.go
tianon-sso: Is that something you'd like me to update the PR for/with, or just something you want to note? :bow:

elezar: It was initially just a note from my side, but what about adding an `ensureDevice(*oci.Spec, *Device)` function that handles all of this internally?

elezar: (similarly for `ensureMount` where we follow the same pattern). Not blockers from my side though.

pkg/cdi/container-edits.go
elezar: A named loop is fine! Thanks.

klihub:
> A first pass of this looks good. One question I would have is whether we wanted to add a local type that wrapps `*oci.Spec` like `runtime-tools` does to reduce the diff further, but I assume that it probably isn't worth the effort for the reduced feature set we have here.
> 
> @klihub what do you think?

@elezar I think that would be cleaner.

pkg/cdi/container-edits.go
```diff
@@ -80,9 +79,11 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
 		return nil
 	}
 
-	specgen := ocigen.NewFromSpec(spec)
 	if len(e.Env) > 0 {
-		specgen.AddMultipleProcessEnv(e.Env)
+		if spec.Process == nil {
```
klihub, line 83: Maybe this could be an `ensureProcess()` for consistency.

pkg/cdi/container-edits.go
```diff
@@ -115,80 +119,131 @@ func (e *ContainerEdits) Apply(spec *oci.Spec) error {
...
 	}
 
-	if len(e.NetDevices) > 0 {
-		// specgen is currently missing functionality to set Linux NetDevices,
-		// so we use a locally rolled function for now.
-		for _, dev := range e.NetDevices {
-			specgenAddLinuxNetDevice(&specgen, dev.HostInterfaceName, (&LinuxNetDevice{dev}).toOCI())
-		}
+	for _, dev := range e.NetDevices {
+		ensureLinuxNetDevices(spec)
```
klihub, line 136: If we kept the length check it would be more consistent and we could call `ensureLinuxNetDevices()` once in the if branch outside the for loop.

klihub (approved):
I have a few minor nits. 

@elezar I am fine with this if you and others consider eliminating our dependency on runtime-tools that important. I am not sure and I tend to think that a properly maintained runtime-tools/generate would be more useful than re-implementing the same, in this case arguably small, functionality multiple times wherever runtime-tools/generate could be used, but this is just my subjective opinion.

elezar:
> A first pass of this looks good. One question I would have is whether we wanted to add a local type that wrapps `*oci.Spec` like `runtime-tools` does to reduce the diff further, but I assume that it probably isn't worth the effort for the reduced feature set we have here.

I agree that a well-maintained `runtime-tools` would be better for the community as a whole, but I don't know whether that is something that we have control over.

One thing to note is that at least for `moby/moby` the dependency on `runtime-tools` was introduced with the addition of CDI in https://github.com/moby/moby/pull/45134. Even for containerd, it was re-added in https://github.com/containerd/containerd/pull/6654 (after being added and then removed in https://github.com/containerd/containerd/pull/2097 and https://github.com/containerd/containerd/pull/3169, respectively).

Both Podman and Cri-o have been using the dependency from the beginning.

pkg/cdi/container-edits.go
elezar: See also https://github.com/opencontainers/runtime-tools/pull/809

elezar:
@tianon-sso thanks again for this change. We decided to go with #317 instead as it more clearly leaves open the option to switch back to `runtime-tools/generate` at a later stage.

PR #317, opened 2026-06-15
Remove runtime-tools dependency

elezar:
Removes runtime-tools by replacing generator calls with an internal OCI spec editor. This allows us to also switch back to using a `Generator` from `runtime-tools/generate` if we feel that this makes sense at some stage. Note that a similar interface is used in https://github.com/containerd/containerd/pull/13519.

Includes the env cache keying fix from opencontainers/runtime-tools#808 and focused tests.

Assisted by Codex


elezar:
@tianon-sso thanks for the work on #316. This here aims to reduce the diff by introducing a `SpecEditor` interface similar to what you did for container.

@klihub, this *should* also allow us to more easily switch back to `runtime-tools` if required at some stage.

klihub (approved):
@elezar Thank you ! LGTM.

internal/ociedit/spec_editor.go
```diff
@@ -0,0 +1,247 @@
...
+	"slices"
+	"strings"
+
+	oci "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// SpecEditor is the internal boundary between CDI edits and OCI spec mutation.
+// It deliberately models only the operations CDI needs, so a future external
+// generator can be wired in without leaking that dependency elsewhere.
+type SpecEditor interface {
```
klihub, line 30: @elezar One thing that came to mind is that, since one of the ideas of the newly introduced interface in this PR is to spell out an explicit interface we use for editing OCI Specs and to do it with a runtime-tools/generate.Generator-compatible fashion, maybe we could have a test case which simply asserts that runtime-tools/generate.Generator indeed implements our SpecEditor interface.

elezar: The issue with this is that we would then once again need to import the dependency -- which would go against the actual intent of this PR and #316. (We could add it in one of the `cmds` so as to keep the dependency focussed there).

klihub (approved):

tianon-sso:
For extra context, `runtime-tools` is effectively unmaintained (and IMO should be officially deprecated and archived but I don't get to make that call in a vacuum) and I wouldn't recommend switching back to it for any reason.

PR #321, opened 2026-07-12
parser: guard single-character vendor and class names against a slice-bounds panic

### LINKED ISSUE #320: parser: single-character vendor or class name causes a slice-bounds panic
What happened:

`validateVendorOrClassName` in `pkg/parser/parser.go` slices `name[1 : len(name)-1]` without a guard for `len(name) == 1`. For a single-character name that expression is `name[1:0]`, whose low bound exceeds its high bound, so Go panics with `slice bounds out of range [1:0]`.

`ValidateVendorName("a")` and `ValidateClassName("b")` both panic. Normal multi-character names return a nil error as expected.

The sibling function `ValidateDeviceName` does the same style of slicing but guards it with `if len(name) == 1 { return nil }`. The vendor/class validator is missing that one guard, so the difference looks like an oversight rather than a deliberate rejection of single-character names.

Two public entry points reach the validator:

1. `ParseAnnotations` calls `parser.IsQualifiedName` for each CDI device name, which calls `ParseQualifiedName` and then `ValidateVendorName` / `ValidateClassName`. `cdi.ParseAnnotations(map[string]string{"cdi.k8s.io/x": "a/b=c"})` panics rather than returning an error.
2. On-disk spec validation reaches the same validator: `Spec.validate` (`pkg/cdi/spec.go:216`) calls `parser.ValidateVendorName` and `parser.ValidateClassName` (spec.go:220 and 223). A spec file with a single-character vendor or class kind panics when the cache loads it.

Reachability:

An ordinary Kubernetes pod cannot trigger this directly. Pod annotations are not forwarded to the CRI as container annotations, so a `cdi.k8s.io/*` pod annotation does not reach `ParseAnnotations`. The sources that can supply a single-character name are a device plugin or DRA driver that emits a `cdi.k8s.io/` container annotation, or a spec file under a CDI spec directory such as `/etc/cdi`. So this is robustness and availability hardening rather than a tenant-facing issue.

Impact:

The library ships no `recover()` on this path. Container runtimes that vendor CDI (containerd and CRI-O both vendor v1.1.0 and call `ParseAnnotations` during container creation) do not wrap this call in a local recovery, so the panic can terminate the calling runtime process unless an outer recovery boundary intercepts it. A device-injection library should not panic on malformed input.

Suggested fix:

Add a guard in `validateVendorOrClassName`, right after the first-character letter check, the same one `ValidateDeviceName` already has:

```go
if len(name) == 1 {
    return nil
}
```

A single letter already satisfies both the first-character letter rule and the final-character alphanumeric rule, so returning nil for length 1 is correct and does not loosen validation.

thc1006:
Fixes #320

## What

Add `if len(name) == 1 { return nil }` to `validateVendorOrClassName`, the guard `ValidateDeviceName` already has.

## Why

`validateVendorOrClassName` slices `name[1 : len(name)-1]`, which is `name[1:0]` for a single-character name and panics with `slice bounds out of range [1:0]`. The value reaches this code through `ParseAnnotations` and through on-disk `Spec.validate`, and the library has no recovery on that path, so a malformed name can crash the calling runtime process.

## Tests

Added single-character vendor and class cases to `TestQualifiedName`; these panicked before the fix. With the guard applied, a 45-second fuzz run of about 9.6 million executions over `ParseQualifiedName` found no other crash on this path.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -150,6 +150,9 @@ func validateVendorOrClassName(name string) error {
 	if !IsLetter(rune(name[0])) {
 		return fmt.Errorf("%q, should start with letter", name)
 	}
+	if len(name) == 1 {
+		return nil
+	}
 	for _, c := range string(name[1 : len(name)-1]) {
 		switch {
 		case IsAlphaNumeric(c):
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -41,6 +41,31 @@ func TestQualifiedName(t *testing.T) {
 			isQualified: true,
 		},
 		{
+			// single-character vendor and class (regression: used to panic
+			// with slice bounds out of range [1:0])
+			device:      "a/b=c",
+			vendor:      "a",
+			class:       "b",
+			name:        "c",
+			isQualified: true,
+		},
+		{
+			// single-character class
+			device:      "aa/b=cc",
+			vendor:      "aa",
+			class:       "b",
+			name:        "cc",
+			isQualified: true,
+		},
+		{
+			// single-character vendor
+			device:      "a/bb=cc",
+			vendor:      "a",
+			class:       "bb",
+			name:        "cc",
+			isQualified: true,
+		},
+		{
 			device:      "vendor.com/class=0",
 			vendor:      "vendor.com",
 			class:       "class",


pkg/parser/parser.go
```diff
@@ -150,6 +150,13 @@ func validateVendorOrClassName(name string) error {
 	if !IsLetter(rune(name[0])) {
 		return fmt.Errorf("%q, should start with letter", name)
 	}
+	if len(name) == 1 {
+		// A single letter is a valid vendor/class name: it already satisfies
+		// both the leading-letter and trailing-alphanumeric rules. Without this
+		// guard the slice below is name[1:0], which panics. This mirrors the
+		// len(name)==1 guard in ValidateDeviceName.
```
elezar: My only nit is that this comment is redundant. We don't have a similar one in `ValidateDeviceName`.

thc1006: Good point, dropped the comment to match `ValidateDeviceName`. Thanks!

klihub (approved):

marquiz (approved):
LGTM

elezar (approved):

PR #322, opened 2026-07-15
cdi: add resolved edits API

ebendler:
The idea here is that some CDI consumers may not be operating on an OCI container (think sandbox wrappers), but it would be silly to have them all re-implement the code that maps the specfile to the running host when the reference implementations are already doing all of that work.

See also: https://github.com/cncf-tags/container-device-interface-rs/pull/101

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -226,8 +226,6 @@ func (c *Cache) refreshIfRequired(force bool) (bool, error) {
 // any of the devices. Might trigger a cache refresh, in which case any
 // errors encountered can be obtained using GetErrors().
 func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, error) {
-	var unresolved []string
-
 	if ociSpec == nil {
 		return devices, fmt.Errorf("can't inject devices, nil OCI Spec")
 	}
@@ -237,23 +235,8 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
 
 	_, _ = c.refreshIfRequired(false) // we record but ignore errors
 
-	edits := &ContainerEdits{}
-	specs := map[*Spec]struct{}{}
-
-	for _, device := range devices {
-		d := c.devices[device]
-		if d == nil {
-			unresolved = append(unresolved, device)
-			continue
-		}
-		if _, ok := specs[d.GetSpec()]; !ok {
-			specs[d.GetSpec()] = struct{}{}
-			edits.Append(d.GetSpec().edits())
-		}
-		edits.Append(d.edits())
-	}
-
-	if unresolved != nil {
+	edits, unresolved := c.collectContainerEdits(devices)
+	if len(unresolved) > 0 {
 		return unresolved, fmt.Errorf("unresolvable CDI devices %s",
 			strings.Join(unresolved, ", "))
 	}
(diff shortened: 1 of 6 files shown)


pkg/cdi/resolved-edits.go
```diff
@@ -0,0 +1,158 @@
...
+	ContainerPath string
+	Type          string
+	Major         *int64
+	Minor         *int64
+	FileMode      *os.FileMode
+	Permissions   string
+	UID           *uint32
+	GID           *uint32
+	Scope         EditScope
+}
```
klihub: nit: Couldn't we/wouldn't it be more natural to use here (and in ResolvedMount) a struct embedding (or containing a field of) the original type and an additional scope field ?

So something like

```golang
type ResolvedDeviceNode struct {
    DeviceNode
    Scope EditScope
}
```

pkg/cdi/resolved-edits.go
```diff
@@ -0,0 +1,158 @@
...
+
+// ResolvedMount is a CDI mount without OCI conversion or normalization.
+type ResolvedMount struct {
+	HostPath      string
+	ContainerPath string
+	Type          string
+	Options       []string
+	Scope         EditScope
+}
```
klihub: nit: Couldn't we/wouldn't it be more natural to use here a struct embedding (or containing a field of) the original type and an additional scope field ?

So something like

```golang
type ResolvedMount struct {
    Mount
    Scope EditScope
}
```

ebendler:
Thanks, @klihub — your comments made the result API clearer. I’ve updated the resolved device-node and mount outputs to embed the effective CDI types with scope metadata, and added coverage for the FIFO zero-value convention.

pkg/cdi/resolved-edits.go
```diff
@@ -0,0 +1,152 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cdi
+
+import cdi "tags.cncf.io/container-device-interface/specs-go"
+
+// ResolvedEdits contains CDI edits resolved without applying OCI semantics.
+type ResolvedEdits struct {
```
elezar: My first thought here is why we need a new type and can't just return the appended `ContainerEdits`? I'm not saying that distinguishing between supported and unsupported edits is not useful, but I would argue that this is something that the CLIENT must decide and not be hard-coded into the CDI API implementation.

ebendler: I think the new version is more what you had in mind?

pkg/cdi/cache.go
```diff
@@ -237,34 +235,78 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
 	if err := edits.Apply(ociSpec); err != nil {
 		return nil, fmt.Errorf("failed to inject devices: %w", err)
 	}
 
 	return nil, nil
 }
 
+// ResolveEdits resolves the given qualified devices to their effective CDI
+// edits without applying OCI conversion or mutating an OCI Spec.
+func (c *Cache) ResolveEdits(devices ...string) (*ContainerEdits, error) {
```
elezar, line 266: Question: Should this return a `cdi.ContainerEdits` type instead?

pkg/cdi/cache.go
```diff
@@ -237,34 +235,78 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+	}
+
+	resolved := cloneContainerEdits(nil)
+	for _, entry := range selected {
+		copy := cloneContainerEdits(entry)
+		if err := copy.resolveDeviceNodes(); err != nil {
+			return nil, err
+		}
+		resolved.Append(copy)
+	}
```
elezar: Instead of manually implementing clone functions that would need to be updated for every added non-value field, what about constructing a copy by marshalling to JSON and then unmarshalling again?

elezar: What about:
1. call `collectContainerEdits`
2. Use the same flow to construct a single `ContainerEdits` that we use in `InjectDevices`
3. Marshal the collected edits to JSON. 
4. Unmarshal the selected edits.
5. Fill missing info for included device nodes.

This also seems to indicate that `collectContainerEdits` should be:
```
func (c *Cache) collectContainerEdits(devices []string) (*ContainerEdits, error) {
	...
}
```
and include its current content as well as unresolved error handling and looping over the resolved edits.

pkg/cdi/cache.go
```diff
@@ -237,34 +235,78 @@ func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, e
...
+
+	return resolved, nil
+}
```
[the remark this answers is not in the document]
elezar: We could decide whether we want to release the lock after we have cloned the collected ContainerEdits.

pkg/cdi/resolved-edits.go
a5ehren: Implemented in 0eacf94: ResolveEdits now returns a complete *ContainerEdits aggregate with no supported/unsupported classification, so consumers retain their own support-policy decisions.

pkg/cdi/cache.go
a5ehren: Implemented in 0eacf94: collectContainerEdits now builds and returns one *ContainerEdits aggregate, which InjectDevices consumes directly and ResolveEdits snapshots.

pkg/cdi/cache.go
a5ehren: Implemented in 0eacf94: replaced the per-field clone helpers with JSON marshal/unmarshal of the embedded specs-go ContainerEdits.

a5ehren: Implemented in 0eacf94: collectContainerEdits now builds the shared aggregate; ResolveEdits clones that aggregate while locked, then fills missing device-node metadata on the detached result.

pkg/cdi/cache.go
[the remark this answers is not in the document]
a5ehren: Implemented in 0eacf94: the cache lock now covers refresh, selection, and snapshot creation only. resolveDeviceNodes runs after the detached snapshot has been created and the lock released.

pkg/cdi/container-edits.go
```diff
@@ -70,6 +71,39 @@ type ContainerEdits struct {
 	*cdi.ContainerEdits
 }
 
+// cloneContainerEdits returns a deep copy of edits that can be safely modified
+// without changing the source CDI Spec.
+func cloneContainerEdits(edits *ContainerEdits) (*ContainerEdits, error) {
```
klihub, line 76: nit: why not `func (edits *ContainerEdits) Clone() (*ContainerEdits, error)` ?

pkg/cdi/cache.go
```diff
@@ -239,8 +239,6 @@ func (c *Cache) refreshIfRequired(force bool) (bool, error) {
 // any of the devices. Might trigger a cache refresh, in which case any
 // errors encountered can be obtained using GetErrors().
 func (c *Cache) InjectDevices(ociSpec *oci.Spec, devices ...string) ([]string, error) {
-	var unresolved []string
```
klihub, line 242: I think this will conflict with #345, so if we're to take both in we should co-ordinate and rebase either one on the other. Due to their nature, I think it'll be easier to take #345 in first, then rebase this one.

PR #325, opened 2026-07-16
cmd/cdi: respect the --spec-dirs flag

marquiz:
Also, don't bail out on spec dir errors – behave similarly to default spec dirs.

### DIFF
--- a/cmd/cdi/cmd/root.go
+++ b/cmd/cdi/cmd/root.go
@@ -67,15 +67,8 @@ func initSpecDirs() {
 	cdi.SetSpecValidator(schema.WithSchema(s))
 
 	if len(specDirs) > 0 {
-		cache, err := cdi.NewCache(
-			cdi.WithSpecDirs(specDirs...),
-		)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/cdi/cmd/root.go
```diff
@@ -67,15 +67,8 @@ func initSpecDirs() {
...
-		cache, err := cdi.NewCache(
-			cdi.WithSpecDirs(specDirs...),
-		)
-		if err != nil {
-			fmt.Printf("failed to create CDI cache: %v\n", err)
-			os.Exit(1)
-		}
-		if len(cache.GetErrors()) > 0 {
-			cdiPrintCacheErrors()
+		if err := cdi.Configure(cdi.WithSpecDirs(specDirs...)); err != nil {
```
klihub, line 70: Do we now still print any cache errors after this change ? If not then we should readd/do it here IMO (without bailing out on errors).

marquiz: Depends on the sub-command. For `validate`, yes we do.

But the main idea I had is, work the same way independent whether we use the default spec dirs or user-specified spec dirs.

PR #329, opened 2026-08-07
schema: make validation error private

thaJeztah:
- relates to https://github.com/cncf-tags/container-device-interface/pull/54

The exported Error type exposes gojsonschema.Result, leaking details of the JSON schema validator implementation, as well as the dependency itself, into the public API.

Error was introduced in 711b138f81fa71144c3dcf76ed78f8fb957a90da as part of PR 54. Neither the commit nor the related PR discussion indicate an intent for the validator result or the Error type itself to form part of the public API; it appears to have only been used internally to carry validation details until the error was formatted.

There are no known external consumers of the schema package at all, let alone consumers depending on the exported Error type or its Result field.

Make the validation error implementation private. This keeps the public API independent of the JSON schema validator and allows that implementation and dependency to be replaced without affecting callers.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -58,9 +58,9 @@ func (s *Schema) Validate(spec *cdi.Spec) error {
 	return s.ValidateType(spec)
 }
 
-// Error wraps a JSON validation result.
-type Error struct {
-	Result *schema.Result
+// validationError wraps a JSON validation result.
+type validationError struct {
+	result *schema.Result
 }
 
 // Set sets the default validating JSON schema.
@@ -239,7 +239,7 @@ func (s *Schema) validate(doc schema.JSONLoader) error {
 		return nil
 	}
 
-	return &Error{Result: docErr}
+	return &validationError{result: docErr}
 }
 
 type schemaContents map[string]interface{}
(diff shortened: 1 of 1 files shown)


thaJeztah:
@elezar @klihub PTAL; I was looking at possible replacements for the unmaintained json-schema dependencies, when I noticed that the dependency leaked into the public API of the schema module (yikes!).

From a search though, it looks like there's no external consumers of the schema module, and the Error type (and Result) field were only used internally within the package, so I think we should be fine to un-export this.

- https://pkg.go.dev/tags.cncf.io/container-device-interface/schema?tab=importedby
- https://grep.app/search?q=%22tags.cncf.io%2Fcontainer-device-interface%2Fschema%22
- GitHub search (only CDI itself and forks); https://github.com/search?q=%22tags.cncf.io%2Fcontainer-device-interface%2Fschema%22&type=code

klihub:
> @elezar @klihub PTAL; I was looking at possible replacements for the unmaintained json-schema dependencies...

Yes, it would be nice to find a properly maintained alternative.

klihub (approved):
Well, if your summary is an accurate description of the current situation, then I don't see what this could break and how... so LGTM.

thaJeztah:
Thanks! Yes, I started working on a branch to switch to https://github.com/santhosh-tekuri/jsonschema, which looks to be the implementation that I see many projects converge on; it has very little dependencies, which is also nice;

```
require (
	github.com/dlclark/regexp2 v1.11.0 // used for testing
	golang.org/x/text v0.14.0
)
```

thaJeztah:
@elezar @klihub pushed my branch (only the last commit is the actual migration / rewrite);

- https://github.com/cncf-tags/container-device-interface/pull/331

elezar (approved):

PR #330, opened 2026-08-07
schema, validation: minor fixes and modernization

thaJeztah:
### chore: go fix code

Automated modernizing code with `go fix -mod=readonly -all ./...`

### schema: implement BuiltinSchema with sync.OnceValue

Replace the hand-crafted lazy initialization of the built-in schema with
sync.OnceValue.

This simplifies the implementation and makes concurrent calls to
BuiltinSchema safe without requiring explicit synchronization.

### schema: synchronize access to the active schema

Set can replace the active schema while it is being used for validation,
causing a data race on current.

Use an atomic.Pointer for current, and use Get for reads, so changing the
active schema is safe while validations are running concurrently.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -28,6 +28,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"sigs.k8s.io/yaml"
 
@@ -35,6 +36,13 @@ import (
 	"tags.cncf.io/container-device-interface/internal/validation"
 	cdi "tags.cncf.io/container-device-interface/specs-go"
 )
+
+// currently loaded schema, builtin by default
+var current atomic.Pointer[Schema]
+
+func init() {
+	current.Store(BuiltinSchema())
+}
 
 const (
 	// BuiltinSchemaName names the builtin schema for Load()/Set().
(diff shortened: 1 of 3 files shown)


schema/schema.go
```diff
@@ -36,6 +37,13 @@ import (
 	cdi "tags.cncf.io/container-device-interface/specs-go"
 )
 
+// currently loaded schema, builtin by default
+var current atomic.Pointer[Schema]
+
+func init() {
+	current.Store(BuiltinSchema())
+}
```
elezar, line 45: If I recall correctly, we removed the implicit loading of the `BuiltinSchema` at some point due to the performance implications of this. Have we since addressed these by not importing the schema sub-module by default?

schema/schema.go
```diff
@@ -352,12 +350,8 @@ func (e *Error) Error() string {
...
 }
 
-var (
-	// our builtin schema
-	builtin *Schema
-	// currently loaded schema, builtin by default
-	current = BuiltinSchema()
-)
+// currently loaded schema, builtin by default
+var current = builtinSchema()
```
elezar, line 354: Ah: I see now, we were aready always loading it.

elezar (approved):

elezar:
@klihub did you also want to take a look at this one?

klihub (approved):
LGTM

klihub:
> @klihub did you also want to take a look at this one?

Yes, LGTM and approved.

PR #333, opened 2026-08-07
specs-go: remove x/mod semver dependency

thaJeztah:
- [x] stacked on https://github.com/cncf-tags/container-device-interface/pull/332

----

### specs-go: remove x/mod semver dependency

CDI spec versions are a small, fixed set, so using
`golang.org/x/mod/semver` just to compare them is unnecessary.

Keep the versions in order instead and use that ordering for validation and
minimum-version checks. This also keeps support for versions with or without a
leading `v`.

As a side effect, `ValidateVersion` gets noticeably faster for valid versions
in local benchmarks, with no change in allocations.

I wrote a quick benchmark, but probably not worth to keep (so I can drop that commit);

```
pkg: tags.cncf.io/container-device-interface/specs-go
                          │ before.txt  │              after.txt              │
                          │   sec/op    │   sec/op     vs base                │
ValidateVersion/first-2     276.9n ± 1%   155.8n ± 1%  -43.71% (p=0.000 n=10)
ValidateVersion/middle-2    279.4n ± 1%   171.1n ± 1%  -38.77% (p=0.000 n=10)
ValidateVersion/last-2      354.9n ± 0%   259.9n ± 1%  -26.78% (p=0.000 n=10)
ValidateVersion/invalid-2   131.9n ± 1%   137.9n ± 1%   +4.55% (p=0.000 n=10)
geomean                     245.3n        175.8n       -28.33%

                          │ before.txt │              after.txt              │
                          │    B/op    │    B/op     vs base                 │
ValidateVersion/first-2     592.0 ± 0%   592.0 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/middle-2    592.0 ± 0%   592.0 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/last-2      672.0 ± 0%   672.0 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/invalid-2   280.0 ± 0%   280.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                     506.7        506.7       +0.00%
¹ all samples are equal

                          │ before.txt │              after.txt              │
                          │ allocs/op  │ allocs/op   vs base                 │
ValidateVersion/first-2     5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/middle-2    5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/last-2      8.000 ± 0%   8.000 ± 0%       ~ (p=1.000 n=10) ¹
ValidateVersion/invalid-2   4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                     5.318        5.318       +0.00%
¹ all samples are equal
```

### DIFF
--- a/specs-go/version_test.go
+++ b/specs-go/version_test.go
@@ -1,0 +1,252 @@
+/*
+   Copyright © The CDI Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package specs_test
+
+import (
+	"strings"
+	"testing"
+
+	"tags.cncf.io/container-device-interface/specs-go"
+)
+
+func TestMinimumRequiredVersion(t *testing.T) {
+	tests := []struct {
+		doc  string
+		spec *specs.Spec
+		want string
+	}{
+		{
+			doc:  "no version-specific features",
+			spec: &specs.Spec{},
+			want: "0.3.0",
+		},
+		{
+			doc: "v0.4.0 mount type",
+			spec: &specs.Spec{
+				ContainerEdits: specs.ContainerEdits{
+					Mounts: []*specs.Mount{{Type: "bind"}},
+				},
+			},
+			want: "0.4.0",
+		},
+		{
+			doc: "v0.5.0 device node host path",
+			spec: &specs.Spec{
+				ContainerEdits: specs.ContainerEdits{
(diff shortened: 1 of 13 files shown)


specs-go/version_test.go
```diff
@@ -0,0 +1,252 @@
...
+package specs_test
+
+import (
+	"strings"
+	"testing"
+
+	"tags.cncf.io/container-device-interface/specs-go"
+)
+
+func TestMinimumRequiredVersion(t *testing.T) {
```
elezar, line 26: A nit: There is already `TestRequiredVersion` that overlaps with the tests here. I'm happy to move them though.

elezar (changes requested):

specs-go/version_test.go
thaJeztah: OH! Just realised this was part of the other PR, which was already merged;

- https://github.com/cncf-tags/container-device-interface/pull/332

Happy to (re)move things around though!

klihub (approved):

elezar (approved):

PR #335, opened 2026-08-08
pkg/cdi: stop embedding Cache mutex

thaJeztah:
The cache mutex has been embedded since the initial cache implementation in b730453b8e7e4550ebd7c7b7e836b5f8e64929c0, unintentionally promoting `Lock` and `Unlock` as part of the public `Cache` API.

Move the mutex to an unexported `mu` field and add deprecated forwarding methods for compatibility. This keeps existing callers working while making cache locking an implementation detail and allows the synchronization primitive to change independently in the future.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -37,7 +37,7 @@ type Option func(*Cache)
 
 // Cache stores CDI Specs loaded from Spec directories.
 type Cache struct {
-	sync.Mutex
+	mu        sync.Mutex
 	specDirs  []string
 	specs     map[string][]*Spec
 	devices   map[string]*Device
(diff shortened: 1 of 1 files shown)


pkg/cdi/cache.go
```diff
@@ -48,6 +48,20 @@ type Cache struct {
...
+func (c *Cache) Lock() {
+	c.mu.Lock()
+}
+
+// Unlock unlocks the cache.
+//
+// Deprecated: Cache locking is an implementation detail and should not be managed by callers.
+func (c *Cache) Unlock() {
+	c.mu.Unlock()
+}
```
thaJeztah, line 63: Not sure how likely it is someone is messing with the mutex; added the deprecated wrappers to keep SemVer compatibility, but perhaps it's something we can still wiggle out of.

cc @klihub @elezar 

(was dusting off a branch I had with various changes in this package, and splitting it up into sizeable commits).

elezar: I think it's unlikely that callers are locking / unlocking the cache directly, but marking these as deprecated should at least call out when that is the case.

I'd also be Ok with just removing them, but we should probably avoid that now that we're v1.x.

elezar (approved):

PR #336, opened 2026-08-08
replace strings.Split for strings.Cut

thaJeztah:
Use strings.Cut to simplify the code and to reduce allocations. Also simplify annotation checks for readability.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -83,13 +83,12 @@ func ParseDevice(device string) (string, string, string) {
 		return "", "", device
 	}
 
-	parts := strings.SplitN(device, "=", 2)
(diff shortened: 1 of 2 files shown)


marquiz (approved):
Good improvement, thx @thaJeztah 

One minor nit: contains unrelated change (`for` -> `if`), could be at least mentioned in the commit message

thaJeztah:
Ah, yes, I had a branch with a bunch of changes that I started to work on some time ago, and trying to slice-and-dice it into small PRs.

Updated; while looking at it, also updated to use `if len(x) > 0` instead of `if len(x) != 0`, which I think is more readable.

marquiz (approved):
LGTM

elezar (approved):

PR #339, opened 2026-08-09
pkg/cdi: deprecate ErrStopScan and use os.ReadDir for scanning Spec directories

thaJeztah:
### pkg/cdi: deprecate ErrStopScan

`ErrStopScan` was introduced with the internal `scanSpecFunc` callback in
b730453b8e7e4550ebd7c7b7e836b5f8e64929c0, but never used, and external
callers have never been able to use it to control Spec scanning.

Deprecate it, and remove special handling from the scanner.

### pkg/cdi: use os.ReadDir for scanning Spec directories

Spec directory scanning is intentionally non-recursive, so use `os.ReadDir`
instead of `filepath.Walk` and process only the immediate directory entries.

This better matches the intended behavior, simplifies the scanning logic,
and avoids the `os.Lstat` call that `filepath.Walk` performs for every file
or directory.

### pkg/cdi: Cache.refresh: remove redundant filepath.Clean

The function is used as a callback for scanSpecDirs, which constructs
the paths passed to the callback with `filepath.Join`, and already cleaned.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -172,7 +172,6 @@ func (c *Cache) refresh() error {
 	}
 
 	_ = scanSpecDirs(c.specDirs, func(path string, priority int, spec *Spec, err error) error {
-		path = filepath.Clean(path)
 		if err != nil {
 			collectError(fmt.Errorf("failed to load CDI Spec %w", err), path)
 			return nil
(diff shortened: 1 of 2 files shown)


elezar (approved):

klihub (approved):
LGTM

PR #340, opened 2026-08-09
remove "sigs.k8s.io/yaml" dependency

thaJeztah:
- [x] depends on / stacked on https://github.com/cncf-tags/container-device-interface/pull/328

### remove "sigs.k8s.io/yaml" dependency

Use go.yaml.in/yaml/v3 for both marshaling and unmarshaling CDI specs,
and drop the sigs.k8s.io/yaml dependency.

The Kubernetes YAML implementation converts YAML to JSON before decoding,
which is useful for Kubernetes APIs where JSON semantics are part of the API
contract. CDI does not require that intermediate representation, and its spec
types already provide YAML tags matching their JSON representation.

Use Decoder.KnownFields(true) to retain strict rejection of unknown fields.
yaml.v3 also rejects duplicate mapping keys, preserving the strictness relied
on from sigs.k8s.io/yaml.UnmarshalStrict.

Keep the existing behavior of decoding only the first YAML document rather
than introducing additional validation for trailing documents.

### DIFF
--- a/cmd/cdi/cmd/inject.go
+++ b/cmd/cdi/cmd/inject.go
@@ -22,7 +22,7 @@ import (
 	"os"
 
 	oci "github.com/opencontainers/runtime-spec/specs-go"
-	"sigs.k8s.io/yaml"
+	"go.yaml.in/yaml/v3"
 
 	"github.com/spf13/cobra"
 )
--- a/cmd/cdi/go.mod
+++ b/cmd/cdi/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/spf13/cobra v1.6.0
 	github.com/stretchr/testify v1.12.1
 	go.yaml.in/yaml/v3 v3.0.5
-	sigs.k8s.io/yaml v1.4.0
 	tags.cncf.io/container-device-interface v1.1.0
 	tags.cncf.io/container-device-interface/schema v0.0.0
 	tags.cncf.io/container-device-interface/specs-go v1.1.0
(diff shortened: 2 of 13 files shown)


klihub:
@thaJeztah If go.yaml.in/yaml/v3 has stable unmarshalling order (of maps) which IIUC it does, than this LGTM. IIRC, that was the primary/only reason at the time to switch to using sigs.k8s.io/yaml.

thaJeztah:
Hm, right; we should check; tests passed, but not sure if they're testing that behavior?

In the meantime, I rebased https://github.com/cncf-tags/container-device-interface/pull/342 to not be stacked on top of this one (at least that bump drops some of the legacy YAML packages).

klihub:
> Hm, right; we should check; tests passed, but not sure if they're testing that behavior?

I don't think we verify that via test. For marshalling we could have a test for it. For unmarshalling we definitely should not care. It is a user-friendliness preference, not mandated by the specification of the CDI Spec in any way. A CDI Spec is perfectly valid if fields are not in declaration order but it is kosher otherwise.



> 
> In the meantime, I rebased #342 to not be stacked on top of this one (at least that bump drops some of the legacy YAML packages).

@thaJeztah It has exactly the stable, deterministic field marshalling order we prefer: fields are marshalled in field declaration order. It also supports strict unmarshalling, and your patch uses it. So I think this should be good to go in. Feel free to mark this as ready for review, and IMO this should be good to go in.

thaJeztah:
@klihub thanks SO MUCH for investigating; I've hit similar discussions a few times, and I keep forgetting about some of the stable ordering (admitted, the import here gave some clues), but tests passed so ...

Also happy to add a test; I think that's the common thing I also encountered in other repositories, and it also wasn't tested there :joy:


I did a quick rebase to get a fresh run of CI and move it out of draft 👍

klihub:
> @klihub thanks SO MUCH for investigating; I've hit similar discussions a few times, and I keep forgetting about some of the stable ordering (admitted, the import here gave some clues), but tests passed so ...

I'm glad to help any way I can.

> Also happy to add a test; I think that's the common thing I also encountered in other repositories, and it also wasn't tested there 😂
> 
> I did a quick rebase to get a fresh run of CI and move it out of draft 👍

@thaJeztah Since we touched upon CI/tests, could you take a look at #356. It's a fix for a flake you also took a stab at, but it wasn't quite enough to slay the beast.

thaJeztah:
Saw it! Just when I was pushing a fix as well when I found my mistake 😂;

> LOL! I just was about to push something similar 😂 Also found that I overlooked a case where it existed early; then thought "perhaps we should just share the assertions"; let me know which one you like more;
> 
> * [pkg/cdi: TestRefreshCache: consolidate cache state checks #357](https://github.com/cncf-tags/container-device-interface/pull/357)

klihub (approved):

PR #341, opened 2026-08-09
pkg/cdi: remove redundant darwin implementation

thaJeztah:
The darwin implementation was identical to the other unix-implementations.


thaJeztah:
Ah, derp; the non-darwin doesn't have an error return 😂

PR #345, opened 2026-08-20
pkg/cdi: fix, and cleanup some locks, and use RWMutex

thaJeztah:
- part of https://github.com/cncf-tags/container-device-interface/pull/344

### pkg/cdi: reduce lock scope in InjectDevices

InjectDevices only needs to hold the lock while refreshing the cache and
capturing the current devices map. Release the lock before resolving devices
to reduce lock contention while applying the container edits on the passed-in
OCI spec.

### pkg/cdi: newCache: remove redundant locking and simplify

This is a constructor, working on a freshly created `Cache`, so there
should be no concurrency to account for. While updating, also set the
default specDirs directly from a clone of `DefaultSpecDirs`, instead of
applying through `WithSpecDirs` to avoid the extra indirect.

Paths are still cleaned, to account for the `DefaultSpecDirs` being
a mutable, package-level variable (but generally not recommended to
update).

### pkg/cdi: fix error collection in device cache

Rename vars that shadowed imports, and make sure that the returned errors
are properly dereferenced. Also append directory errors instead of overwriting
existing errors for the same path.

### DIFF
--- a/pkg/cdi/cache.go
+++ b/pkg/cdi/cache.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


thaJeztah:
cc @marquiz @klihub ptal 🤗

klihub (approved):
LGTM

PR #353, opened 2026-08-21
specs-go: fix benchmark compatibility with go1.19

thaJeztah:
The go.mod defines go1.19, which does not yet support b.Loop;

    # tags.cncf.io/container-device-interface/specs-go [tags.cncf.io/container-device-interface/specs-go.test]
    Error: ./version_bench_test.go:19:10: b.Loop undefined (type *testing.B has no field or method Loop)
    FAIL	tags.cncf.io/container-device-interface/specs-go [build failed]

### DIFF
--- a/specs-go/version_bench_test.go
+++ b/specs-go/version_bench_test.go
@@ -16,7 +16,7 @@ func BenchmarkValidateVersion(b *testing.B) {
 	for _, tc := range tests {
 		b.Run(tc.doc, func(b *testing.B) {
 			b.ReportAllocs()
-			for b.Loop() {
+			for i := 0; i < b.N; i++ {
 				_ = ValidateVersion(&Spec{Version: tc.version})
 			}
 		})


specs-go/version_bench_test.go
```diff
@@ -16,7 +16,7 @@ func BenchmarkValidateVersion(b *testing.B) {
 	for _, tc := range tests {
 		b.Run(tc.doc, func(b *testing.B) {
 			b.ReportAllocs()
-			for b.Loop() {
+			for i := 0; i < b.N; i++ {
```
bart0sh, line 19: nit:
```suggestion
			for range b.N {
```

thaJeztah: That requires go1.22 or up, and the go.mod still defines go1.19 as minimum

elezar (approved):
@bart0sh I think @thaJeztah's explaination is sufficient for using an explicit for loop.

Please merge if you have no other blockers.

elezar:
> Please merge if you have no other blockers.

Actually. Considering the nature of this test-only change, I'm going to merge it to unblock @thaJeztah.

PR #356, opened 2026-08-25
cache test: fix up flaky self refresh test.

klihub:
Fix flaky self refresh test/corner case.

/cc @thaJeztah

### DIFF
--- a/pkg/cdi/cache_test.go
+++ b/pkg/cdi/cache_test.go
@@ -235,6 +235,7 @@ func TestRefreshCache(t *testing.T) {
 		errors  []map[string]struct{}
 		devices [][]string
 		devprio []map[string]int
+		waitSrc string
 	}
 	for _, tc := range []*testCase{
 		{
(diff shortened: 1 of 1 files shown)


thaJeztah:
LOL! I just was about to push something similar 😂 Also found that I overlooked a case where it existed early; then thought "perhaps we should just share the assertions";  let me know which one you like more;

- https://github.com/cncf-tags/container-device-interface/pull/357

klihub:
> LOL! I just was about to push something similar 😂 Also found that I overlooked a case where it existed early; then thought "perhaps we should just share the assertions"; let me know which one you like more;
> 
> * [pkg/cdi: TestRefreshCache: consolidate cache state checks #357](https://github.com/cncf-tags/container-device-interface/pull/357)

@thaJeztah I think yours makes more sense. I was going for bare minimal changes for the fix without touching much else. Yours restructures the verification in a way that IMO makes more sense, since with that there is no more this slightly weird split of 'let's verify a/some bits, so we can decide to move on and verify more'. So let's go with yours.

PR #357, opened 2026-08-25
pkg/cdi: TestRefreshCache: consolidate cache state checks

thaJeztah:
Run all cache state checks through a single EventuallyWithT assertion instead of duplicating a subset for automatic refreshes.

The previous wait only checked device names and errors. When a new Spec shadowed an existing device without changing either, those conditions were already satisfied, allowing the wait to complete before the watcher had updated the cached device priority.

Include all state checks in the wait so both manual and automatic refreshes use the same assertions and watcher-driven updates cannot be observed prematurely.

updates b6e2ef921f9a3caf9eeb815c311220bf5ab670b4

### DIFF
--- a/pkg/cdi/cache_test.go
+++ b/pkg/cdi/cache_test.go
@@ -562,12 +562,7 @@ devices:
 							t.Errorf("failed to update test directory: %v", err)
 							return
 						}
-						if selfRefresh {
-							require.EventuallyWithT(t, func(t *assert.CollectT) {
-								assert.Equal(t, tc.devices[idx], cache.ListDevices())
-								assert.Len(t, cache.GetErrors(), len(tc.errors[idx]))
-							}, time.Second, 10*time.Millisecond)
-						} else {
+						if !selfRefresh {
 							err = cache.Refresh()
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
⚠️ 👉 easier to review with indentation changes hidden; add `?w=1` to the diff URL; https://github.com/cncf-tags/container-device-interface/pull/357/changes?w=1

pkg/cdi/cache_test.go
```diff
@@ -578,27 +573,32 @@ devices:
 						}
 					}
 
-					devices := cache.ListDevices()
-					if len(tc.devices[idx]) == 0 {
-						require.True(t, len(devices) == 0)
-					} else {
-						require.Equal(t, tc.devices[idx], devices)
-					}
+					require.EventuallyWithT(t, func(t *assert.CollectT) {
```
thaJeztah, line 576: "eventually" for the manual-refresh case will just be "immediately" (but it may retry for max 1 second, which probably is OK)

klihub (approved):
LGTM. Let's pick/merge this instead of #356 (which I then close).

PR #359, opened 2026-08-27
chore(deps): github.com/fsnotify/fsnotify v1.7.0

thaJeztah:
Update to the lowest version that uses tagged versions of golang.org/x/sys to reduce transitive dependencies. Most projects would already be on v1.9 or v1.10, so this is still very conservative.


elezar (approved):
LGTM. 

Can you think of any reason to not bump this to a more recent release in a follow-up?

PR #361, opened 2026-09-04
Bump version to v1.1.1

elezar:
This includes a Spec module bump to v1.1.1.

Note that since there are no spec changes, we keep v1.1.0 as the current Spec version. The go module version is bumped to (and will be tagged as) `v1.1.1` though.
 
Related to #360

### DIFF
--- a/schema/go.mod
+++ b/schema/go.mod
@@ -6,8 +6,8 @@ require (
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.3
 	github.com/stretchr/testify v1.12.1
 	go.yaml.in/yaml/v3 v3.0.5
-	tags.cncf.io/container-device-interface v1.1.0
-	tags.cncf.io/container-device-interface/specs-go v1.1.0
+	tags.cncf.io/container-device-interface v1.1.1
+	tags.cncf.io/container-device-interface/specs-go v1.1.1
 )
 
 require (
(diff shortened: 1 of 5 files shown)


elezar:
cc @thaJeztah

klihub (approved):
LGTM.

schema/go.mod
thaJeztah, line 23: FWIW, it's good to remove the replace rules (temporarily), or instead use a `go.work` for this; ISTR some go commands don't like a replace rule present;

```bash
go install tags.cncf.io/container-device-interface/cmd/cdi@main
go: downloading tags.cncf.io/container-device-interface/cmd/cdi v0.0.0-20260828125525-73444d1f71f2
go: downloading tags.cncf.io/container-device-interface v1.1.1-0.20260828125525-73444d1f71f2
go: tags.cncf.io/container-device-interface/cmd/cdi@main (in tags.cncf.io/container-device-interface/cmd/cdi@v0.0.0-20260828125525-73444d1f71f2):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

thaJeztah (approved):
LGTM on tagging a new release, but left a comment (we can temporarily remove the replace rules, then add them back in a follow-up, or look at using `go.work`)

schema/go.mod
elezar: What do you mean by replacing them temporarily? I'd have to read up on migrating to `go.work` at some point.

Do you mean remove them here and then cut the release? This means that this mod file is only valid ONCE tags have been created since the referenced versions don't exist.

elezar: Also, for what it's worth. The `cmd`s are not really consumed as a stable API. They were added early on to provide some utility, but I'm not aware of anyone that actually `go installs` them.

PR #362, opened 2026-09-08
Support wildcard device paths

marquiz:
Allows the `path` and `hostPath` of a `deviceNodes` entry to be a wildcard pattern, so a device can cover all instances of a class of device nodes without the vendor having to enumerate them. An example:

```yaml
cdiVersion: "1.2.0"
kind: "vendor.com/device"
devices:
  - name: "all"
    containerEdits:
      deviceNodes:
        - path: "/dev/dri/card*"
        - path: "/dev/mei[1-9]"
          permissions: "rw"
```

Patterns use [`path.Match`](https://pkg.go.dev/path#Match) syntax. Each match becomes a device node
of its own. Non-device matches (regular files, directories, symlinks) are ignored, and a pattern matching nothing expands to an empty list. The `type`, `major` and `minor` fields are inherited from the host device node and cannot be specified.

Also bumps the spec version to v1.2.0.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -8,7 +8,7 @@
 
 ## Version
 
-This is CDI **spec** version **1.1.0**.
+This is CDI **spec** version **1.2.0**.
 
 ### Update policy
 
@@ -35,6 +35,7 @@ Released versions of the spec are available as Git tags.
 | v1.0.0 |   | Move minimum version logic to specs-go package. |
 | v1.1.0 |   | Add `NetDevices` to `ContainerEdits`, `Schemata` and `EnableMonitoring` to `IntelRdt`. Dropped `EnableCMT` and `EnableMBM` fields from `IntelRdt`. |
 | v1.1.1 |   | No Spec changes. Remove semver dependency from specs-go and improve performance. |
+| v1.2.0 |   | Allow wildcards in the `Path` and `HostPath` of `DeviceNodes`. |
 
 *Note*: spec loading fails on unknown fields and when the minimum required version is higher than the version specified in the spec. The minimum required version is determined based on the usage of fields mentioned in the table above. For example the minimum required version is v0.6.0 if the `Annotations` field is used in the spec, but `IntelRdt` is not.
 `MinimumRequiredVersion` API can be used to get the minimum required version.
@@ -129,6 +130,7 @@ The keywords "must", "must not", "required", "shall", "shall not", "should", "sh
             ]
             "deviceNodes": [ (optional)
                 {
+                    // The last element of the path may be a wildcard pattern which gets expanded into the matching host device nodes.
                     "path": "<path>",
                     "hostPath": "<hostPath>" (optional),
                     "type": "<type>" (optional),
@@ -230,8 +232,8 @@ The `containerEdits` field is referenced in two places in the specification:
 The `containerEdits` field has the following definition:
   * `env` (array of strings in the format of "VARNAME=VARVALUE", OPTIONAL) describes the environment variables that should be set. These values are appended to the container environment array.
   * `deviceNodes` (array of objects, OPTIONAL) describes the device nodes that should be mounted:
-    * `path` (string, REQUIRED) path of the device within the container.
-    * `hostPath` (string, OPTIONAL) path of the device node on the host. If not specified the value for `path` is used. Added in v0.5.0.
+    * `path` (string, REQUIRED) path of the device within the container. May be a wildcard pattern, see [Wildcard device nodes](#wildcard-device-nodes). Wildcards added in v1.2.0.
+    * `hostPath` (string, OPTIONAL) path of the device node on the host. If not specified the value for `path` is used. May be a wildcard pattern, see [Wildcard device nodes](#wildcard-device-nodes). Added in v0.5.0, wildcards added in v1.2.0.
     * `type` (string, OPTIONAL) Device type: block, char, etc.
     * `major` (int64, OPTIONAL) Device major number.
     * `minor` (int64, OPTIONAL) Device minor number.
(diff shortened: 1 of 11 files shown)


eero-t (commented):
> The `type`, `major` and `minor` fields are inherited from the host device node and cannot be specified.

Wouldn't that break `device_ownership_from_security_context` option: https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/ ?

pkg/cdi/container-edits_unix.go
```diff
@@ -76,6 +78,120 @@ func deviceInfoFromPath(path string) (*deviceInfo, error) {
...
+	matches, err := filepath.Glob(pattern)
+	if err != nil {
+		// NOTE: the pattern is checked by Validate() so we should never end up here
+		return nil, fmt.Errorf("invalid device node pattern %q: %w", pattern, err)
+	}
+
+	expanded := make([]*cdi.DeviceNode, 0, len(matches))
+	for _, hostPath := range matches {
+		// Ignores matches which are not device nodes, also filtering out symlinks (like /dev/dri/by-path/*)
+		if _, err := deviceInfoFromPath(hostPath); err != nil {
```
klihub, line 177: @marquiz Can we really assume here that the only possible reason for failure is `hostPath` having been expanded to a non-device entry by wildcard match ? It is not obvious to me why we could assume that blindly, and if we get any other error here it is now silently dropped.

If my understanding is correct, then I think it would be better to codify and verify the assumption here and fail on any other errors. I'd suggest
- adding a `var global ErrNotADevice = errors.New("not a device node")`
- returning that error from `deviceInfoFromPath()`, and 
- updating the match-ignoring check here in `expand()` to something like

```golang
...
        if _, err := deviceInfoFromPath(hostPath); err != nil {
            if errors.Is(err, ErrNotADevice) {
                continue
            }
            return nil, fmt.Errorf("failed to fill device info for CDI host device %q: %w", hostPath, err)
        }
....
```

klihub (commented):
@marquiz This looks good to me apart from something that at first glance looks to me like an overly permissive assumption about errors related to filtering glob-matched non-device paths. Can you take a look at the more detailed related comment ?

elezar:
@marquiz under which cases would a vendor not want to enumerate devices? This seems like a stop-gap for dynamic behaviour that we should address differently. As @klihub points out, there is a concern that this will end up being overly permissive. It also means that a CDI spec does not define what is actually made available to a container. Here what is actually injected depends on the system state at the point of consumption and this may lead to uninteded behaviour.

specs-go/wildcard.go
```diff
@@ -0,0 +1,31 @@
...
+
+const wildcardMetaChars = `*?[`
+
+// HasWildcards reports whether the given device node path is a wildcard
+// pattern.
+//
+// NOTE: a path containing a meta character is always treated as a pattern,
+// even if the meta character is escaped with a backslash. Escaping only affects
+// what the pattern matches, not whether the path is a pattern.
+func HasWildcards(path string) bool {
```
elezar, line 29: Did we want to expose this as a public API? (We don't do this for any other validation).

pkg/cdi/container-edits_unix.go
```diff
@@ -76,6 +78,120 @@ func deviceInfoFromPath(path string) (*deviceInfo, error) {
...
+		if !cdi.HasWildcards(p) {
+			continue
+		}
+		if _, err := filepath.Match(p, ""); err != nil {
+			return fmt.Errorf("device %q: invalid wildcard pattern %q: %w", d.Path, p, err)
+		}
+		// Require absolute path, and a "clean" path (rule out paths like "/dev/../etc/*")
+		if !filepath.IsAbs(p) || filepath.Clean(p) != p {
+			return fmt.Errorf("device %q: wildcard pattern %q is not an absolute, cleaned path", d.Path, p)
+		}
```
elezar, line 95: Should we not check this BEFORE we glob?

SPEC.md
```diff
@@ -262,6 +264,43 @@ The `containerEdits` field has the following definition:
...
+* A pattern MUST be an absolute path with no `.` or `..` elements.
+* A pattern is expanded when the container edits are applied to an OCI
+  specification, i.e. against the state of the host at container creation time.
+* A pattern is expanded to the matching host device nodes only, any other files
+  (regular files, directories etc) are silently ignored.
+* A pattern matching no host device nodes expands to an empty set.
+* `type`, `major` and `minor` MUST NOT be specified for a wildcard device
+  node, these are inherited from the host device node.
+* If `hostPath` is not specified, the pattern is matched against the host
+  filesystem. In this case wildcards may appear in any element of the path.
```
elezar: Why do we allow wildcards in any path element?

specs-go/wildcard.go
marquiz: This is re-used by `pkg/cdi` so that's the reason.

pkg/cdi/container-edits_unix.go
marquiz: Either works for me. It's not globbing against anything real, just checking the pattern is correctly formatted. I'll move this check first if you think it's important

SPEC.md
marquiz: That was kinda "automatic", keeping it generic. But we don't really need that for any actual usage scenario in mind so I'll make it stricter.

pkg/cdi/container-edits_unix.go
marquiz: Yeah, the comment is a bit off. TBH, I'm not entirely sure what would be the correct action regarding other than not-a-device-node errors (basically lstat fails for some reason). I thought about this during the implementation and figured that maybe skipping is better than failing container restart. No problem changing this but would be good to hear what others think.

Thoughts @elezar @kad @bart0sh ?

marquiz:
> under which cases would a vendor not want to enumerate devices? This seems like a stop-gap for dynamic behaviour that we should address differently. As @klihub points out, there is a concern that this will end up being overly permissive. It also means that a CDI spec does not define what is actually made available to a container. Here what is actually injected depends on the system state at the point of consumption and this may lead to uninteded behaviour.

The idea/motivation behind this was that a container restart would "sync" the devices. Not needing to re-create the Pod. Targeting management containers. If this is not going to fly, then #363 can still be used to work around this (by giving the container CAP_MKNOD and let it create the device nodes itself)

I'm open and eager to hear alternative, better, approaches to dynamic behavior.

pkg/cdi/container-edits_unix.go
```diff
@@ -76,6 +78,120 @@ func deviceInfoFromPath(path string) (*deviceInfo, error) {
...
+
+	for _, p := range []string{d.Path, d.HostPath} {
+		if !cdi.HasWildcards(p) {
+			continue
+		}
+		// Require absolute path, and a "clean" path (rule out paths like "/dev/../etc/*")
+		if !filepath.IsAbs(p) || filepath.Clean(p) != p {
+			return fmt.Errorf("device %q: wildcard pattern %q is not an absolute, cleaned path", d.Path, p)
+		}
+		if _, err := filepath.Match(p, ""); err != nil {
```
klihub, line 96: We should use `path.Match` here instead. IIUC, it has much more robust pattern validation, so it would flag anomalous patterns like `/dev/dri/card*[0-9` as an error, whereas filepath.Match does not flag it.

PR #363, opened 2026-09-08
Add support for extra device cgroup rules

marquiz:
Adds a `deviceCgroupRules` field to `containerEdits`, for granting access to a set of devices in the devices cgroup without creating any device nodes with it. The main motivation is hotplug in device management and device monitoring scenarios, where the device numbers that appear on the host after the container was created cannot be known at container creation time. An omitted  minor is a wildcard matching all device numbers. An example:

```yaml
cdiVersion: "1.2.0"
kind: "vendor.com/device"
devices:
  - name: "all"
    containerEdits:
      deviceCgroupRules:
        - type: "c"
          major: 226
          permissions: "rw"
        - type: "b"
          major: 8
          minor: 0
```

Rules are applied as allow entries in [`linux.resources.devices`](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#allowed-device-list) of the OCI runtime spec.

Also bumps the spec version to v1.2.0.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -24,13 +24,14 @@ type Device struct {
 
 // ContainerEdits are edits a container runtime must make to the OCI spec to expose the device.
 type ContainerEdits struct {
-	Env            []string          `json:"env,omitempty"            yaml:"env,omitempty"`
-	DeviceNodes    []*DeviceNode     `json:"deviceNodes,omitempty"    yaml:"deviceNodes,omitempty"`
-	NetDevices     []*LinuxNetDevice `json:"netDevices,omitempty"     yaml:"netDevices,omitempty"` // Added in v1.1.0
-	Hooks          []*Hook           `json:"hooks,omitempty"          yaml:"hooks,omitempty"`
-	Mounts         []*Mount          `json:"mounts,omitempty"         yaml:"mounts,omitempty"`
-	IntelRdt       *IntelRdt         `json:"intelRdt,omitempty"       yaml:"intelRdt,omitempty"`       // Added in v0.7.0
-	AdditionalGIDs []uint32          `json:"additionalGids,omitempty" yaml:"additionalGids,omitempty"` // Added in v0.7.0
+	Env               []string            `json:"env,omitempty"               yaml:"env,omitempty"`
+	DeviceNodes       []*DeviceNode       `json:"deviceNodes,omitempty"       yaml:"deviceNodes,omitempty"`
+	DeviceCgroupRules []*DeviceCgroupRule `json:"deviceCgroupRules,omitempty" yaml:"deviceCgroupRules,omitempty"` // Added in v1.2.0
+	NetDevices        []*LinuxNetDevice   `json:"netDevices,omitempty"        yaml:"netDevices,omitempty"`        // Added in v1.1.0
+	Hooks             []*Hook             `json:"hooks,omitempty"             yaml:"hooks,omitempty"`
+	Mounts            []*Mount            `json:"mounts,omitempty"            yaml:"mounts,omitempty"`
+	IntelRdt          *IntelRdt           `json:"intelRdt,omitempty"          yaml:"intelRdt,omitempty"`       // Added in v0.7.0
+	AdditionalGIDs    []uint32            `json:"additionalGids,omitempty"    yaml:"additionalGids,omitempty"` // Added in v0.7.0
 }
 
 // DeviceNode represents a device node that needs to be added to the OCI spec.
@@ -44,6 +45,14 @@ type DeviceNode struct {
 	Permissions string       `json:"permissions,omitempty" yaml:"permissions,omitempty"`
 	UID         *uint32      `json:"uid,omitempty"         yaml:"uid,omitempty"`
 	GID         *uint32      `json:"gid,omitempty"         yaml:"gid,omitempty"`
+}
+
+// DeviceCgroupRule represents a rule to be added to the devices cgroup of the container.
+type DeviceCgroupRule struct { // Added in v1.2.0
+	Type        string `json:"type"                  yaml:"type"`
+	Major       *int64 `json:"major"                 yaml:"major"`
+	Minor       *int64 `json:"minor,omitempty"       yaml:"minor,omitempty"`
+	Permissions string `json:"permissions,omitempty" yaml:"permissions,omitempty"`
 }
 
 // Mount represents a mount that needs to be added to the OCI spec.
--- a/specs-go/version.go
+++ b/specs-go/version.go
@@ -22,7 +22,7 @@ import (
 )
 
 // CurrentVersion is the current version of the Spec.
-const CurrentVersion = "1.1.0"
+const CurrentVersion = "1.2.0"
 
 const (
 	// These represent the released versions of the CDI specification
@@ -36,6 +36,7 @@ const (
 	v080 version = "0.8.0"
 	v100 version = "1.0.0"
 	v110 version = "1.1.0"
+	v120 version = "1.2.0"
 
 	// vEarliest is the earliest supported version of the CDI specification
 	vEarliest version = v030
@@ -49,6 +50,7 @@ var validSpecVersions = []struct {
 	version    version
 	isRequired requiredFunc
 }{
+	{v120, requiresV120},
 	{v110, requiresV110},
 	{v100, requiresV100},
 	{v080, requiresV080},
@@ -126,6 +128,23 @@ func versionIndex(v version) int {
 		}
 	}
 	return -1
+}
+
+// requiresV120 returns true if the spec uses v1.2.0 features.
+func requiresV120(spec *Spec) bool {
+	// The v1.2.0 spec allows device cgroup rules to be specified at a spec level.
+	if len(spec.ContainerEdits.DeviceCgroupRules) > 0 {
+		return true
+	}
+
+	for _, d := range spec.Devices {
+		// The v1.2.0 spec allows device cgroup rules to be specified at a device level.
+		if len(d.ContainerEdits.DeviceCgroupRules) > 0 {
+			return true
+		}
+	}
+
+	return false
 }
 
 // requiresV110 returns true if the spec uses v1.1.0 features.
--- a/specs-go/version_test.go
+++ b/specs-go/version_test.go
@@ -119,6 +119,15 @@ func TestMinimumRequiredVersion(t *testing.T) {
 			want: "1.1.0",
 		},
 		{
+			doc: "v1.2.0 device cgroup rule",
+			spec: &specs.Spec{
+				ContainerEdits: specs.ContainerEdits{
+					DeviceCgroupRules: []*specs.DeviceCgroupRule{{Type: "c"}},
+				},
+			},
+			want: "1.2.0",
+		},
+		{
 			doc: "device-scoped feature",
 			spec: &specs.Spec{
 				Devices: []specs.Device{{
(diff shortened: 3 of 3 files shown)


pkg/cdi/container-edits.go
```diff
@@ -354,6 +369,50 @@ func (d *DeviceNode) Validate() error {
...
+	*cdi.DeviceCgroupRule
+}
+
+// Validate a CDI Spec DeviceCgroupRule.
+func (r *DeviceCgroupRule) Validate() error {
+	// NOTE: stick with the types of the kernel devices cgroup interface.
+	// The extra types ("p" and "u") allowed in the container config are not accepted.
+	// Also, "a" is denied on purpose as would allow all devices ignoring the
+	// major/minor filtering.  A rule for all devices can be expressed as
+	// separatre "b" and a "c" rules.
```
eero-t:
```suggestion
	// Also, "a" is denied on purpose as that would allow all devices and ignore
	// the major/minor filtering.  A rule for all devices can be expressed as
	// separate "b" and a "c" rules.
```

eero-t (commented):
> An omitted major or minor is a wildcard matching all device numbers

May be safer to start by erroring in this case (and relaxing it only if somebody provides strong enough use-case for it, that cannot be handled reasonable in any other way).

klihub:
> > An omitted major or minor is a wildcard matching all device numbers
> 
> May be safer to start by erroring in this case (and relaxing it only if somebody provides strong enough use-case for it, that cannot be handled reasonable in any other way).

Well, I think we at least must allow omitting the minor number, otherwise this can't really be used for what is stated as the primary use case in the description: "The main motivation is hotplug in device management and device monitoring scenarios, where the device numbers that appear on the host after the container was created cannot be known at container creation time."

@marquiz But I was also wondering if we really need to allow such wide permissions, or if we could instead get along with only allowing the minor number to be omitted. Do we really have a use case for injecting access permissions to all block or character devices ?

marquiz:
> But I was also wondering if we really need to allow such wide permissions, or if we could instead get along with only allowing the minor number to be omitted. Do we really have a use case for injecting access permissions to all block or character devices ?

Yeah, agree. I think allowing "wildcard" on minor only is a sound choice (nobody should have the need to allow everything with one single rule). I'll change the PR

pkg/cdi/container-edits.go
marquiz: Thx. Fixed/changed as suggested

marquiz:
> But I was also wondering if we really need to allow such wide permissions, or if we could instead get along with only allowing the minor number to be omitted. Do we really have a use case for injecting access permissions to all block or character devices ?

Done. Major is now required (and must be > 0)

pkg/cdi/container-edits.go
```diff
@@ -354,6 +369,56 @@ func (d *DeviceNode) Validate() error {
...
+	*cdi.DeviceCgroupRule
+}
+
+// Validate a CDI Spec DeviceCgroupRule.
+func (r *DeviceCgroupRule) Validate() error {
+	// NOTE: stick with the types of the kernel devices cgroup interface.
+	// The extra types ("p" and "u") allowed in the container config are not accepted.
+	// Also, "a" is denied on purpose as that would allow all devices and ignore the
+	// major/minor filtering.  A rule for all devices can be expressed as
+	// separate "b" and a "c" rules.
```
klihub: nit: The last bit of this, 'the rule for matching all (block+char) devices, is not true any more, since we require the major number to be specified.

pkg/cdi/container-edits.go
```diff
@@ -354,6 +369,56 @@ func (d *DeviceNode) Validate() error {
...
+// String returns the rule in the "<type> <major>:<minor> <permissions>" format
+// used by the devices cgroup, with "*" denoting a wildcard device number.
+func (r *DeviceCgroupRule) String() string {
+	num := func(n *int64) string {
+		if n == nil {
+			return "*"
+		}
+		return strconv.FormatInt(*n, 10)
+	}
+	return fmt.Sprintf("%s %s:%s %s", r.Type, num(r.Major), num(r.Minor), r.access())
```
klihub, line 418: nit: in principle IIUC this could be now `return fmt.Sprintf("%s %d:%s %s", r.Type, r.Major, num(r.Minor), r.access())`, but it's better to leave this as it is, since the structure itself would allow an unspecified/nil major (which I also think we should leave as it is).

SPEC.md
```diff
@@ -242,6 +256,13 @@ The `containerEdits` field has the following definition:
       * m - allows container to create device files that do not yet exist.
     * `uid` (uint32, OPTIONAL) id of device owner in the container namespace.
     * `gid` (uint32, OPTIONAL) id of device group in the container namespace.
+  * `deviceCgroupRules` (array of objects, OPTIONAL) describes rules to be added to the devices cgroup of the container, see [Device cgroup rules](#device-cgroup-rules). Added in v1.2.0.
+    * `type` (string, REQUIRED) type of the devices the rule applies to, one of:
+      * b - block device.
+      * c - character device.
+    * `major` (int64, REQUIRED) major number of the devices the rule applies to. Must be greater than 0.
+    * `minor` (int64, OPTIONAL) minor number of the devices the rule applies to. If not specified the rule applies to all minor numbers of the given major.
+    * `permissions` (string, OPTIONAL) Cgroups permissions to grant, with the same candidates as the `deviceNodes.permissions`. Omitted or empty default to `rwm`.
```
klihub, line 265: @marquiz @elezar @bart0sh Should we explicitly spell out here, that an actual device injection in the Spec is always an implicit corresponding device cgroup rule injection. IOW, there is absolutely no need to manually specify device cgroup rules for devices you inject by a CDI Spec. You only need to specify device cgroup rules, if you do not inject the actual devices using a CDI Spec. This might not otherwise be fully obvious after this addition to the Spec, especially not for hasty readers.

pkg/cdi/container-edits.go
marquiz: True. Dropped this last sentence from the comment

pkg/cdi/container-edits.go
marquiz: Yeah, I didn't change anything. `%d` would feel like an unnecessary optimization here

elezar:
I still need to do a proper review, but this seems to align with what I proposed in https://github.com/cncf-tags/container-device-interface/pull/314. My first question would be whether a separate containerEdits type is preferred over reusing "deviceNodes"? Using a specific type would help for clarity.

specs-go/config.go
```diff
@@ -24,13 +24,14 @@ type Device struct {
...
-	Env            []string          `json:"env,omitempty"            yaml:"env,omitempty"`
-	DeviceNodes    []*DeviceNode     `json:"deviceNodes,omitempty"    yaml:"deviceNodes,omitempty"`
-	NetDevices     []*LinuxNetDevice `json:"netDevices,omitempty"     yaml:"netDevices,omitempty"` // Added in v1.1.0
-	Hooks          []*Hook           `json:"hooks,omitempty"          yaml:"hooks,omitempty"`
-	Mounts         []*Mount          `json:"mounts,omitempty"         yaml:"mounts,omitempty"`
-	IntelRdt       *IntelRdt         `json:"intelRdt,omitempty"       yaml:"intelRdt,omitempty"`       // Added in v0.7.0
-	AdditionalGIDs []uint32          `json:"additionalGids,omitempty" yaml:"additionalGids,omitempty"` // Added in v0.7.0
+	Env               []string            `json:"env,omitempty"               yaml:"env,omitempty"`
+	DeviceNodes       []*DeviceNode       `json:"deviceNodes,omitempty"       yaml:"deviceNodes,omitempty"`
+	DeviceCgroupRules []*DeviceCgroupRule `json:"deviceCgroupRules,omitempty" yaml:"deviceCgroupRules,omitempty"` // Added in v1.2.0
```
elezar, line 29: As far as I know these are Linux only? Should we name them as such? (we do so for the `LinuxNetDevices`).

specs-go/config.go
```diff
@@ -46,6 +47,14 @@ type DeviceNode struct {
 	GID         *uint32      `json:"gid,omitempty"         yaml:"gid,omitempty"`
 }
 
+// DeviceCgroupRule represents a rule to be added to the devices cgroup of the container.
+type DeviceCgroupRule struct { // Added in v1.2.0
+	Type        string `json:"type"                  yaml:"type"`
+	Major       *int64 `json:"major"                 yaml:"major"`
+	Minor       *int64 `json:"minor,omitempty"       yaml:"minor,omitempty"`
+	Permissions string `json:"permissions,omitempty" yaml:"permissions,omitempty"`
+}
```
elezar, line 56: Question: Should this be embedded in the `DeviceNode` struct?

specs-go/config.go
marquiz: Hmm, good point. "Linux" is probably is justified here. I'll change this

specs-go/config.go
marquiz: I thought about that but my conclusion was no. This is kinda orthogonal feature (IMHO). DeviceNodes creates device nodes. These don't create any device nodes, but give access.

But I'm open to hear arguments against this choice.

klihub: If I understand the suggestion/question correctly, then I think one con against embedding is that it might, despite of any documentation updates we add, give more strongly the wrong impression that one would need to manually specify a [Linux]DeviceCgroupRule for an ordinary device injection.

Another con is that, if we embed, I think it becomes syntactically possible to specify the injection of a single device and as a side-effect providing access to a whole family/class of devices, even unrelated ones. Do we want to support that ? Probably not. So then we'd need to disable/forbid this by validation. Which is much less explicit than not allowing it syntactically in the first place.

klihub:
> I still need to do a proper review, but this seems to align with what I proposed in #314. My first question would be whether a separate containerEdits type is preferred over reusing "deviceNodes"? Using a specific type would help for clarity.

@elezar I think that is the major advantage of this PR's approach. It is rather explicit about what is happening so what users could expect. The wildcard approach is quite a bit more implicit. For instance, without (reading the) documentation some users might not get a clear idea whether any device nodes get injected. To some intuition might give the wrong impression that since its a globbing pattern, matching nodes will be injected. This approach does not suffer from that problem. It would also leave the door open for any future wildcard-based matching device injection if we'd ever to decide to consider adding something like that.

specs-go/config.go
elezar: I was thinking of an anonymous embedding:
```
type DeviceNode struct {
    DeviceCgroupRule
    ...
}
```
to allow for reusing the logic to add the rules later, but it may not be as useful as what I expect.

SPEC.md
```diff
@@ -146,6 +147,19 @@ The keywords "must", "must not", "required", "shall", "shall not", "should", "sh
...
                 }
             ]
+            // Rules to add to the devices cgroup of the container.
+            "deviceCgroupRules": [ (optional)
+                {
+                    // Device type ("b" or "c")
+                    "type": "<type>",
+                    "major": <int64>,
+                    // An omitted or empty minor is a wildcard matching any minor number.
+                    "minor": <int64> (optional),
```
bart0sh, line 157: This is a bit implicit. Would mandatory minor supporting wildcards be a safer option?

marquiz: Dunno :man_shrugging:, what others think. The "wildcard" would be `-1` then

specs-go/config.go
marquiz: I think this still feels non-coherent/illogical, mixing two separate things. Like allows specifying a device node embedding a device cgroup rule that has nothing to do with the device node specified...

SPEC.md
elezar: In order to collect data, how does Docker or Podman handle this?

specs-go/config.go
elezar: Let's not embed it then.

SPEC.md
elezar: Yes, I think making it clear that on systems where cgroups are used, a device node implies a corresponding cgroup rule.

SPEC.md
bart0sh: My rationale FWIW: If minor is unintentionally not added, all minors would be implicitly allowed and everything would work with wider permissions, allowing access to all the minors. This could cause security issues and after all this is not what user actually wanted.

elezar: Yes, I understand. I suppose the issue is that communicating what `-1` means may not be so straightforward.

